PR #5, opened 2021-05-28
First working version of client

kevincasey1222:
Hi @mknoedel  and @aiwilliams  !

Here is a way early version of the GitHub integration rewrite. It runs and can access data, but the tests don't pass. I'm opening this PR as a start to a conversation. It's not done.

A couple of bits of input I'd like from you guys:

1) The test code refers to a test/helpers/polly file that I don't have. Could I get a copy to see what exactly it is doing? I know it's recording stuff - but if I can see specifically how it is mutating the entries, I can make sure we match that in the new SDK.

2) There are a few places where `process.env.SOME_THING` is hardcoded into a function. As I understand it, that won't work in the new SDK when it pulls from a client config. Is that correct? I'd love to know what the names of the config variables will be when passed from client configs, so I can do the same magic we did for Bitbucket to get everything aligned, and move those hard-coded process variable functions to point to the instance config.

Thanks!
Kevin


src/approval/collectCommitsForPR.test.ts
```diff
@@ -0,0 +1,206 @@
...
+      'e42f6281ec53e1611a0b45c012333af688b85d2d'
+    ),
+    approvedCommits: [],
+    commitsByUnknownAuthor: [],
+    approvals: [],
+  },
+  teamMembers: [{ login: 'fomentia2' }],
+});
+
+// test('pull request with commits from author outside of team', async () => {
```
mknoedel, line 122: Does this no longer work?

src/approval/extractPRsFromCommits.ts
```diff
@@ -0,0 +1,25 @@
+import { ReposListCommitsResponseItem } from '../types';
+
+const REF_REGEX = /pull request #([\d]+)/;
+
+function detectPRReference(comment: string): number | undefined {
+  const match = REF_REGEX.exec(comment);
+
+  if (!match) {
+    return undefined;
```
mknoedel, line 9: I think adding a log that a PR couldn't be extracted from a commit would be good.

src/approval/extractPRsFromCommits.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  return Number.parseInt(match[1], 10);
```
mknoedel, line 12: It would be nice to make sure match[1] is a number as well

src/approval/collectCommitsForPR.ts
```diff
@@ -0,0 +1,159 @@
...
+  );
+  return validApprovers.length > 0;
+}
+
+function userOutsideOfTeam(login: string, teamMembers: string[]) {
+  return !teamMembers.includes(login);
+}
+
+function hasTeamApprovals(approval: Approval, teamMembers?: string[]) {
+  if (!teamMembers) {
```
mknoedel, line 111: I think this check should be pulled out of this method and we should only call `hasTeamApprovals` if there are team members. the parameter should also not be optional.

src/client.ts
```diff
@@ -41,41 +48,49 @@ export type AcmeGroup = Opaque<any, 'AcmeGroup'>;
...
-        },
-        (res) => {
-          if (res.statusCode !== 200) {
-            reject(new Error('Provider authentication failed'));
-          } else {
-            resolve();
-          }
-        },
+    const installationId = Number(this.config.installationId);
+    const appClient = await createGitHubAppClient(
```
mknoedel, line 58: 👍

package.json
```diff
@@ -25,12 +25,33 @@
...
   "devDependencies": {
-    "@jupiterone/integration-sdk-core": "^6.0.0",
-    "@jupiterone/integration-sdk-dev-tools": "^6.0.0",
-    "@jupiterone/integration-sdk-testing": "^6.0.0"
+    "@jupiterone/integration-sdk-core": "^6.3.0",
+    "@jupiterone/integration-sdk-dev-tools": "^6.3.0",
+    "@jupiterone/integration-sdk-testing": "^6.3.0",
+    "@octokit/rest": "^17.6.0",
+    "@octokit/auth-app": "2.4.5",
+    "@octokit/plugin-throttling": "^3.4.1."
```
mknoedel, line 55: probably don't need these as both dev and normal dependencies.

package.json
```diff
@@ -25,12 +25,33 @@
...
+    "@octokit/plugin-throttling": "^3.4.1",
+    "@octokit/core": "^3.0.0"
   },
   "devDependencies": {
-    "@jupiterone/integration-sdk-core": "^6.0.0",
-    "@jupiterone/integration-sdk-dev-tools": "^6.0.0",
-    "@jupiterone/integration-sdk-testing": "^6.0.0"
+    "@jupiterone/integration-sdk-core": "^6.3.0",
+    "@jupiterone/integration-sdk-dev-tools": "^6.3.0",
+    "@jupiterone/integration-sdk-testing": "^6.3.0",
```
mknoedel, line 52: Looks like octokit doesn't do typing files the way everyone else does, but they do have some type files which are likely worth a look.

https://www.npmjs.com/package/@octokit/types

src/client.ts
```diff
@@ -41,41 +48,49 @@ export type AcmeGroup = Opaque<any, 'AcmeGroup'>;
...
+      installationId, 
+      this.logger
+    );
+    const { token } = (await appClient.auth({ type: 'installation' })) as {
```
mknoedel, line 62: would be nice to get types with this call if possible

src/client.ts
```diff
@@ -41,41 +48,49 @@ export type AcmeGroup = Opaque<any, 'AcmeGroup'>;
...
+        token,
+        resourceMetadataMap(),
+        this.logger
+      ),
+      logger: this.logger,
+      analyzeCommitApproval: this.config.analyzeCommitApproval,
+    });
+
+    const output = await accountClient.getAccount();
+    console.log(output);
```
mknoedel, line 90: remove console.logs

mknoedel: might also want to verify response

src/client.ts
```diff
@@ -41,41 +48,49 @@ export type AcmeGroup = Opaque<any, 'AcmeGroup'>;
...
+    const output2 = await accountClient.getMembers();
```
mknoedel, line 91: This call is likely too heavy for verify authentication. Is there a way to limit to just one member?

src/client.ts
```diff
@@ -41,41 +48,49 @@ export type AcmeGroup = Opaque<any, 'AcmeGroup'>;
...
+    console.log(output2);
+    
   }
 
   /**
```
mknoedel, line 96: Along the same lines as in the Bitbucket integration, it would be nice to verify the scopes of the token that is returned here and send back a message if they do not have the permissions to ingest parts of this integration

package.json
```diff
@@ -25,12 +25,33 @@
...
+    "qs": "^6.9.1",
+    "uuid": "^3.0.0"
+  },
   "peerDependencies": {
-    "@jupiterone/integration-sdk-core": "^6.0.0"
+    "@jupiterone/integration-sdk-core": "^6.3.0",
+    "@octokit/rest": "~17.6.0",
+    "@octokit/auth-app": "2.4.5",
+    "@octokit/plugin-throttling": "^3.4.1",
+    "@octokit/core": "^3.0.0"
```
mknoedel, line 47: In another PR, it would be nice to update these to their latest versions. Not now though.

src/client/GraphQLClient/client.ts
```diff
@@ -0,0 +1,123 @@
...
+          }
+        }
+      }
+
+      queryCursors = mapResponseCursorsForQuery(pageCursors, queryCursors);
+      queryResources = mapResponseResourcesForQuery(
+        pageCursors,
+        this.resourceMetadataMap,
+        selectedResources
+      ) as OrganizationResource[];
```
mknoedel, line 114: I believe these should be moved to the start of the while loop as they are only used in the while loop and the placement here broke my brain for a bit

Plus this code is mad complex 🙃

src/client/GraphQLClient/response.ts
```diff
@@ -0,0 +1,222 @@
...
+
+export function extractSelectedResources(
+  selectedResources: OrganizationResource[],
+  resourceMetadataMap: ResourceMap<ResourceMetadata>,
+  data: any
+) {
+  const resources: ResourceMap<any> = {};
+  const cursors: ResourceMap<CursorHierarchy> = {};
+
+  extractSelectedResourceFromData(
```
mknoedel, line 82: We need to write this in such a way that it returns the resources and the cursors instead of modifies them, as it is really hard to follow as is.

src/client/GraphQLClient/client.ts
```diff
@@ -0,0 +1,123 @@
...
+
+      const rateLimit = response.rateLimit;
+      rateLimitConsumed += rateLimit.cost;
+
+      const organizationData = response.organization;
+
+      const {
+        resources: pageResources,
+        cursors: pageCursors,
+      } = extractSelectedResources(
```
mknoedel, line 80: After digging into this stuff for a while, this code is basically just at library for managing nested GraphGL queries with multiple levels of cursors. Really complex stuff but really good stuff! We should definitely work on pulling out this code into a library that we can use in other integrations. Plus it would clean up the code a good bit so it won't take so long to review next time 😅

src/client/OrganizationAccountClient.test.ts
```diff
@@ -0,0 +1,156 @@
...
+  test('getTeamMembers', async () => {
+    await expect(client.getTeamMembers()).resolves.toEqual([]);
+  });
+
+  test('getRepositories', async () => {
+    await expect(client.getRepositories()).resolves.toEqual([]);
+  });
+
+  test('getTeamRepositories', async () => {
+    await expect(client.getTeamRepositories()).resolves.toEqual([]);
```
mknoedel, line 64: We need to add a passing test for `getTeamRepositories` and probably the rest of these as well as it has failed in the past.

docs/jupiterone.md
```diff
@@ -1,31 +1,31 @@
...
 
-TODO: Iterate requirements for setting up the integration. Consider the
-following examples:
+- GitHub uses a GitHub App to secure access to their API. Although GitHub Apps
+  exist on personal accounts, the API only supports access from an App installed
+  on an organization account. Therefore, you will need to be a member of a
+  GitHub organization, and have permission to request Apps to be installed for
+  the organization.
+
+TODO: more details here on next steps
```
mknoedel, line 27: Making a note of this TODO for later.

src/client.ts
```diff
@@ -126,24 +119,49 @@ export class APIClient {
...
-    for (const group of groups) {
-      await iteratee(group);
+    if (!this.accountClient) {
+      await this.setupAccountClient();
     }
+    const teams: OrgTeamQueryResponse[] = await this.accountClient.getTeams();
+
+    for (const team of teams) {
+      console.log(team);
+      console.log(`justin case`);
```
mknoedel, line 129: Need to remove these console.logs

src/client.ts
```diff
@@ -126,24 +119,49 @@ export class APIClient {
...
-        ],
-      },
-    ];
-
-    for (const group of groups) {
-      await iteratee(group);
+    if (!this.accountClient) {
+      await this.setupAccountClient();
     }
+    const teams: OrgTeamQueryResponse[] = await this.accountClient.getTeams();
```
mknoedel, line 125: Is there a reason why we need to fetch the teams in `verifiyAuthentication`? If we do need it then we should add a limit 1 perhaps?

src/client.ts
```diff
@@ -126,24 +119,49 @@ export class APIClient {
...
+      console.log(team);
+      console.log(`justin case`);
+      //await iteratee(team);
+    }
+  }
+
+  public async setupAccountClient(): Promise<void> {
+    const installationId = Number(this.config.installationId);
+    const appClient = await createGitHubAppClient(installationId, this.logger);
+    const { token } = (await appClient.auth({ type: 'installation' })) as {
```
mknoedel, line 137: Are we concerned with errors here at all?

src/client/GraphQLClient/response.ts
```diff
@@ -0,0 +1,222 @@
+import {
```
mknoedel, line 1: This whole file should be a part of that GraphQl multi-cursor package mentioned below.

mknoedel: (Unimportant but nice to have) Before that happens though, it would be nice to add some comments and graphQL documentation links to help understand what is going on here. It is pretty difficult to parse as is for me.

Here is a good link on the documentation
https://graphql.org/learn/pagination/#complete-connection-model

src/client/GraphQLClient/types/client.ts
```diff
@@ -0,0 +1,83 @@
+export enum OrganizationResource {
```
mknoedel, line 1: Unimportant, but it would be good to move the `as { token: string }` to here as an auth response as well.

src/client/OrganizationAccountClient.test.ts
```diff
@@ -0,0 +1,156 @@
+import OrganizationAccountClient from './OrganizationAccountClient';
```
mknoedel, line 1: Change this file name to something that it implies that these are only error tests. Something like `OrganizationAccountClientErrorHandling.test`

src/config.ts
```diff
@@ -21,12 +21,17 @@ import { createAPIClient } from './client';
...
 export const instanceConfigFields: IntegrationInstanceConfigFieldMap = {
-  clientId: {
+  githubAppId: {
     type: 'string',
   },
-  clientSecret: {
+  githubAppLocalPrivateKeyPath: {
+    type: 'string',
+  },
+  githubAppLocalCallbackUrl: {
```
mknoedel, line 30: Is this used at all in this integration or is this just used to get the Private Key originally?

src/util/types.ts
```diff
@@ -0,0 +1,9 @@
+export interface JsonWebToken {
+  exp: number;
+  iat: number;
+  iss: any;
+}
+
+export interface SignedJsonWebToken extends JsonWebToken {
```
mknoedel, line 7: Oh, you already have these types. These should be used way back in the Authentication calls then.

src/util/toTime.ts
```diff
@@ -0,0 +1,5 @@
+export default function toTime(
+  time: Date | string | undefined
+): number | undefined {
+  return time ? new Date(time).getTime() : undefined;
```
mknoedel, line 4: It would be nice to check if `new Date(time).getTime()` is `NaN` prior to returning

src/config.ts
```diff
@@ -21,12 +21,17 @@ import { createAPIClient } from './client';
  * `instance.config` in a UI.
  */
 export const instanceConfigFields: IntegrationInstanceConfigFieldMap = {
-  clientId: {
+  githubAppId: {
     type: 'string',
   },
-  clientSecret: {
+  githubAppLocalPrivateKeyPath: {
```
mknoedel, line 27: a

mknoedel, line 27: We should change this to just be `privateKey`. I think Adam might have talked to you about how we sometimes do pre-processing of config fields on JupiterOne servers prior to executing this integration? This would be one of those times. The privateKey extraction logic would be done on the server and this integration should assume that it is given the key string.

It looks like we are going to have a problem in this integration where the development environment is going to need to do this processing manually. When running the integration in a development environment, check for further environment variables that would not be in the config in order to populate the `privateKey` value in the config. For instance, the .env can remain the same with desiring:
```
GITHUB_APP_LOCAL_PRIVATE_KEY_PATH={YOURPATH}/{YOURFILENAME}.private-key.pem
GITHUB_APP_LOCAL_CALLBACK_URL=https://smee.io/{YOURUNIQUEURL}
```
but the config would be 
```json
{
  appId: {
    type: 'string'
  },
  installationId: {
    type: 'string'
  },
  {
     privateKey: {
      type: 'string' 
    }
  }
}

Comment (mknoedel) [reply src/config.ts:27]: The `GITHUB_APP_LOCAL_CALLBACK_URL` is going to need to be completely independent of the integration execution as it is only needed for setting up the development environment. This should be documented extensively in a `docs/development.md` page

File: src/config.ts (hunk 24 above)
Comment (mknoedel) [reply src/config.ts:30]: It looks like it is only used in order to set up the development environment, so it should be taken out of the config.

File: src/steps/account.ts [hunk 28]
```diff
@@ -41,7 +30,7 @@ export const accountSteps: IntegrationStep<IntegrationConfig>[] = [
     entities: [
       {
         resourceName: 'Account',
-        _type: 'acme_account',
+        _type: 'github_account',
```
mknoedel, line 33: It would be nice to pull out all these type values into a `constants.ts` file similar to [the Azure integration.](https://github.com/JupiterOne/graph-azure/blob/master/src/steps/resource-manager/compute/constants.ts)

src/client/OrganizationAccountClient.ts
```diff
@@ -0,0 +1,557 @@
...
+    );
+    if (this.diffsEqual(diffMergeToMaster.files, diffBranchToMaster.files)) {
+      return true;
+    }
+
+    return false;
+  }
+
+  /*******************************************************
+   * PRIVATES - DON'T LOOK OR I'LL REPORT YOU TO THE CPS *
```
mknoedel, line 525: Probably need to remove this comment 🙃

src/client/OrganizationAccountClient.ts
```diff
@@ -0,0 +1,557 @@
...
+    /**
+     * The login of the GitHub account, a required parameter of some endpoints.
+     */
+    login: string;
+    /**
+     * A GitHub API v3 client configured with necessary authentication, binding it
+     * to a specific GitHub account (organization or user), used for accessing
+     * endpoints listed at https://developer.github.com/v3/apps/available-endpoints/.
+     */
+    restClient: Octokit;
```
mknoedel, line 65: We want to ditch this restClient entirely and only use the graphqlClient if that is possible. That refactor can be done in other PRs.

src/config.ts
mknoedel: It looks like there was an `initializeContext` method that did something like this in the old project. That may be a good pattern to accomplish what I am talking about here.

src/client/OrganizationAccountClient.test.ts
mknoedel: We need tests for this client that test the happy paths as well.

src/client/OrganizationAccountClient.ts
```diff
@@ -0,0 +1,557 @@
...
+          owner: account.login,
+          repo: repo.name,
+          pull_number: id,
+        })
+      ).data;
+
+      this.v3RateLimitConsumed++;
+
+      // We don't check the analyzeCommitApproval option here because it is used
+      // for ingesting specific, single PRs.
```
mknoedel, line 311: It would be nice to have this information move to the top of this file instead of hidden all the way down here.

src/client/OrganizationAccountClient.ts
```diff
@@ -0,0 +1,557 @@
...
+    };
+
+    try {
+      const reviews = (await this.v3.pulls.listReviews(listOptions)).data;
+
+      this.v3RateLimitConsumed++;
+
+      return reviews;
+    } catch (err) {
+      this.logger.info({ err, listOptions }, 'pulls.listReviews failed');
```
mknoedel, line 415: I think we should at least do a `logger.error` here. In future we need to add error handling to this whole Client

src/client/OrganizationAccountClient.ts
```diff
@@ -0,0 +1,557 @@
...
+          base,
+          head,
+        })
+      ).data;
+
+      this.v3RateLimitConsumed++;
+
+      return comparison;
+    } catch (err) {
+      this.logger.info({ err }, 'repos.compareCommits failed');
```
mknoedel, line 474: Same that this should be `logger.error`

src/steps/members.ts
```diff
@@ -0,0 +1,74 @@
...
+}: IntegrationStepExecutionContext<IntegrationConfig>) {
+  const apiClient = createAPIClient(instance.config, logger);
+
+  const accountEntity = (await jobState.getData(
+    DATA_ACCOUNT_ENTITY,
+  )) as AccountEntity;
+
+  await apiClient.iterateMembers(async (member) => {
+    const memberEntity = (await jobState.addEntity(
+      toOrganizationMemberEntity(member),
```
mknoedel, line 31: `toOrganizationMemberEntity` should return a UserEntity so you shouldn't need to use `as UserEntity`. If it does not than it should be changed.

kevincasey1222: `toOrganizationMemberEntity` does return a `UserEntity`, but `jobState.addEntity` returns a generic `Entity`. The cast is needed for `toOrganizationHasMemberRelationship`. Is there a better way to accomplish that?

src/sync/converters.ts
```diff
@@ -0,0 +1,386 @@
...
+    name: data.slug,
+    displayName: data.name,
+    full_name: data.name,
+  };
+  setRawData(teamEntity, { name: 'default', rawData: data });
+  return teamEntity;
+}
+
+export function toRepositoryEntity(data: OrgRepoQueryResponse): RepoEntity {
+  const repoEntity: RepoEntity = {
```
mknoedel, line 82: Update to `createIntegrationEntity` from '@jupiterone/integration-sdk-core';

src/sync/converters.ts
```diff
@@ -0,0 +1,386 @@
...
+    _key: `${userEntity._key}|manages|${installationEntity._key}`,
+    _type: 'github_user_manages_account',
+    _class: 'MANAGES',
+    _fromEntityKey: userEntity._key,
+    _toEntityKey: installationEntity._key,
+    displayName: 'MANAGES',
+  };
+}
+
+export function toOrganizationHasTeamRelationship(
```
mknoedel, line 269: No need for these direct relationship methods anymore, update to `createDirectRelationship` from '@jupiterone/integration-sdk-core' as well.

src/sync/entities/pullRequestContext.ts
```diff
@@ -0,0 +1,94 @@
...
+        pr.id,
+        accumulator.users.length > 0 ? accumulator.users : undefined,
+        Object.keys(accumulator.userByLoginMap).length > 0
+          ? accumulator.userByLoginMap
+          : undefined
+      );
+
+      if (prFromGitHub) {
+        addPullRequest(accumulator, prFromGitHub);
+      }
```
mknoedel, line 42: Add an `logger.info` if a PR was not found.

src/sync/entities/pullRequestContext.ts
```diff
@@ -0,0 +1,94 @@
...
+  );
+}
+
+const pullRequestContext: IngestEntitiesContext<IngestPullRequestEntity> = {
+  loadProviderData: async (executionContext, prs) => {
+    return loadProviderData({
+      github: executionContext.github,
+      logger: executionContext.logger,
+      loadContextSpecificData: async (client, data, logger) => {
+        if (data.account && data.account.accountType === AccountType.Org) {
```
mknoedel, line 54: We only support Organization accounts so I don't think this check is necessary anymore.

src/sync/entities/pullRequestContext.ts
```diff
@@ -0,0 +1,94 @@
...
+          await loadUsersFromGitHub(client, data, logger);
+        }
+        await loadReposFromGitHub(
+          client,
+          data,
+          prs.map(pr => pr.repository)
+        );
+        await loadPullRequests(client, data, prs);
+      },
+      allData: false,
```
mknoedel, line 66: Making a note that we will need to think a bit more if we can load all the data or not, as the old integration had a hack where larger accounts wouldn't load allData but other accounts would.

src/sync/entities/pullRequestContext.ts
```diff
@@ -0,0 +1,94 @@
...
+      ];
+
+      if (executionContext.accountType === AccountType.Org) {
+        loaders.push(
+          loadUserEntitiesFromGraph(graph, data),
+          loadAccountUserRelationshipsFromGraph(graph, data)
+        );
+      }
+
+      await Promise.all(loaders);
```
mknoedel, line 87: How many loaders are we expecting to get here? Do we need to do a pQueue to ensure we don't blow out the memory?

mknoedel: There seems to be a good amount of things getting loaded here.

mknoedel: In fact, I'm not sure this should be used anymore as we are ingesting via the new SDK right now. Should this just get deleted?

src/types/persister.ts
```diff
@@ -0,0 +1,94 @@
+import { AccountType } from '.';
```
mknoedel, line 1: We should no longer know about the "persister". Do we get any value from these types anymore? My gut says we remove these types entirely.

src/util/fetchPrivateKey.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+    if (content) {
+      return content.toString();
+    } else {
+      throw new Error(`${envLocalPathParam} ${localPath}: cannot read content`);
+    }
+  }
+}
+
+async function getSsmPrivateKey(envSssParam: string) {
```
mknoedel, line 26: SSM should not be referenced here as this is only available on the server.  Should be removed.

src/util/fetchPrivateKey.ts
```diff
@@ -0,0 +1,48 @@
+const fs = require('fs');
+//import * as ssm from '../aws/util/ssm';
+
+interface CreateJsonWebTokenOptions {
+  privateKeyEnvLocalPathParam: string;
+  privateKeyEnvSsmParam: string;
+}
+
+function getLocalPrivateKey(envLocalPathParam: string): string | undefined {
```
mknoedel, line 9: This functionality should be moved somewhere out of the src folder as we don't want this bundled in the dist folder.

src/util/fetchPrivateKey.ts
```diff
@@ -0,0 +1,48 @@
+const fs = require('fs');
+//import * as ssm from '../aws/util/ssm';
```
mknoedel, line 2: Should remove this comment.

src/util/isHealthCheck.ts
```diff
@@ -0,0 +1,8 @@
+export const HEALTHY_RESPONSE = {
```
mknoedel, line 1: We do not need this file anymore. It should be deleted.

src/sync/entities/pullRequestContext.ts
```diff
@@ -0,0 +1,94 @@
...
+      if (prFromGitHub) {
+        addPullRequest(accumulator, prFromGitHub);
+      }
+    },
+    { concurrency: 2 }
+  );
+}
+
+const pullRequestContext: IngestEntitiesContext<IngestPullRequestEntity> = {
+  loadProviderData: async (executionContext, prs) => {
```
mknoedel, line 49: We don't need to use loadProviderData anymore as the new sdk takes care of that.

mknoedel: I'm pretty sure `loadProviderData` should be removed.

src/sync/entities/pullRequestContext.ts
```diff
@@ -0,0 +1,94 @@
...
+          data,
+          prs.map(pr => pr.repository)
+        );
+        await loadPullRequests(client, data, prs);
+      },
+      allData: false,
+    });
+  },
+
+  loadGraphData: async (executionContext, prs) => {
```
mknoedel, line 70: I think this is no longer used as the graph ingestion is taken care of in the sdk.

src/types/integration.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+export interface GitHubIntegrationExecutionContext
+  extends IntegrationExecutionContext {
+  //graph: GraphClient;
+  //persister: PersisterClient;
+  accountType: AccountType;
+  github: OrganizationAccountClient;
+}
+
+export interface GraphData {
```
mknoedel, line 65: We will no longer need this Type.

src/util/createGitHubAppClient.ts
```diff
@@ -0,0 +1,69 @@
+import { Octokit } from '@octokit/rest';
+import { createAppAuth } from '@octokit/auth-app';
+import { throttling } from '@octokit/plugin-throttling';
+import { IntegrationLogger } from '@jupiterone/integration-sdk-core';
+
+import fetchPrivateKey from '../util/fetchPrivateKey';
+
+const MAX_RETRIES = 20;
```
mknoedel, line 8: Add a comment that there we don't have a reason for the 20 retry limit.

src/util/createGitHubAppClient.ts
```diff
@@ -0,0 +1,69 @@
...
+   * The auth hook also authenticates requests, manages cached JWTs and tokens,
+   * and authenticates using the installation ID passed to createAppAuth.
+   */
+  const v3 = new OctokitThrottling({
+    userAgent: 'jupiter-integration-github',
+    authStrategy: createAppAuth,
+    // Options passed to authStrategy
+    auth: {
+      id: appId,
+      privateKey: await fetchPrivateKey({
```
mknoedel, line 36: This would turn into a config value, instead of this `fetchPrivateKey` method.

src/util/createGitHubAppClient.ts
```diff
@@ -0,0 +1,69 @@
...
+      },
+      onAbuseLimit: (retryAfter: number, options: any) => {
+        logger.warn(
+          { retryAfter, options },
+          'Abuse limit reached for request.'
+        );
+
+        if (options.request.retryCount < MAX_RETRIES) {
+          return true;
+        }
```
mknoedel, line 63: Because we are not doing error handling here, we will need to make sure that the clients are sending the correct error types when we hit these throttle errors.

src/util/getInstallationId.ts
```diff
@@ -0,0 +1,17 @@
...
+): number {
+  const installationId: any = context.instance.config.installationId;
+
+  if (!installationId) {
+    throw new Error(
+      'No installationId found in integration instance configuration'
+    );
+  }
+
+  return Number(installationId);
```
mknoedel, line 16: Would be nice to check if it is `NaN` first. If so, throw an `IntegrationConfig` error.

src/util/propertyHelpers.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  return collection.reduce((aggregatedProperties: T[], source: any) => {
+    aggregatedProperties.push(source[property]);
+    return aggregatedProperties;
+  }, []);
+}
+
+export function flattenMatrix<T>(matrix: T[][]): T[] {
+  return matrix.reduce((flatArray: T[], row) => {
+    flatArray.push(...row);
```
mknoedel, line 19: it is more efficient to do `return flattenArray.concat(row)` as the splat `...` can get really memory intensive if the row has enough values.

src/util/toTime.test.ts
```diff
@@ -0,0 +1,16 @@
...
+
+test('should return number from Date', () => {
+  const testData: Date = new Date('2019-04-22T21:43:53.000Z');
+  expect(toTime(testData)).toEqual(1555969433000);
+});
+
+test('should return undefined from undefined', () => {
+  const testData = undefined;
+  expect(toTime(testData)).toEqual(undefined);
+});
```
mknoedel, line 16: A `NaN` test would be nice.

PR #7, opened 2021-06-03
Better patterns

kevincasey1222:
Hi @mknoedel !

A lot of work has been done to clean things up, but I've still got a few tests that are not passing due to my struggles to connect to the existing Polly recordings. I wanted to push this up as a PR mainly to have you take a look at two things:

1) Does the new handling of the config look right? (validateInvocation, test/config.ts, etc.)
2) How do I get collectCommitsForPr.test.ts to access the Polly recordings? Same question for `src/client/GraphQLClient/client.test.ts`.

So it's still a work in progress.... Teams and Repos are in this code too, if you want to glance at them. But I still haven't looked at better error handling and packaging inside the REST and GraphQL clients, per your comments in the previous PR.

Kevin


src/approval/__recordings__/collectCommitsForPR-dismissedReview_1084922421/recording.har
```diff
@@ -8,7 +8,7 @@
     },
     "entries": [
       {
-        "_id": "4f8ae50e0d39e502846f5aa0170ac933",
```
mknoedel, line 11: It's ok that these are getting overridden right? I saw that you found a way to use the existing recordings, are you planning on going back and using these?

src/approval/collectCommitsForPR.test.ts
```diff
@@ -30,13 +38,34 @@ function collectCommitsForPRTest({
...
-
-    const executionContext = createTestGitHubIntegrationExecutionContext();
-    const githubContext = await initializeContext(executionContext);
+    p = setupGithubRecording({ directory: __dirname, name: recordingName });
+
+    const context = createMockStepExecutionContext<IntegrationConfig>({
+      instanceConfig: integrationConfig,
+    });
+
+    //mutate config with installation ID 953957, which is used in recordings
```
mknoedel, line 47: Nice!

src/approval/getCommitsToDestination.ts
```diff
@@ -3,15 +3,16 @@ import { PullsListCommitsResponseItem } from '../types';
 function commitMatches(commit: string, match: string): boolean {
   // using the length of the proposed match allows for matching shas of
   // different length, so that abcdefghij12345678 matches abcdefghij
+  //if (!commit) { return false; }
```
mknoedel, line 6: comment

src/approval/__recordings__/collectCommitsForPR-dismissedReview_1084922421/recording.har
kevincasey1222: Yeah. It's non-obvious in the way Github tries to display the change in the PR, but I generated the new guids and then copied in the response from the existing recordings. So the recorded responses are exactly what we had before.

src/approval/getCommitsToDestination.ts
kevincasey1222: Oops. Construction debris during experiments.

src/client.ts
```diff
@@ -97,65 +65,116 @@ export class APIClient {
...
-    // should maintain pagination state, and for each page, for each record in
-    // the page, invoke the `ResourceIteratee`. This will encourage a pattern
-    // where each resource is processed and dropped from memory.
-
     if (!this.accountClient) {
       await this.setupAccountClient();
     }
     const teams: OrgTeamQueryResponse[] = await this.accountClient.getTeams();
-
+    const allTeamMembers: OrgTeamMemberQueryResponse[] = await this.accountClient.getTeamMembers();
```
mknoedel, line 85: Is there another way to get only the team members from a specific team or is this the only way to get this information?

src/client.ts
```diff
@@ -97,65 +65,116 @@ export class APIClient {
...
+    const allTeamRepos: OrgTeamRepoQueryResponse[] = await this.accountClient.getTeamRepositories();
```
mknoedel, line 86: Same with the repos

src/client.ts
```diff
@@ -97,65 +65,116 @@ export class APIClient {
...
     for (const team of teams) {
```
mknoedel, line 87: This is a pretty slow pattern for finding these intersections. It would be nice to take another run at this one to see if we can take out some cyclomatic complexity.

src/client.ts
```diff
@@ -97,65 +65,116 @@ export class APIClient {
...
+   *
+   * @param iteratee receives each resource to produce entities/relationships
+   */
+  public async iterateRepos(
+    iteratee: ResourceIteratee<OrgRepoQueryResponse>,
+  ): Promise<void> {
+    if (!this.accountClient) {
+      await this.setupAccountClient();
+    }
+    const repos: OrgRepoQueryResponse[] = await this.accountClient.getRepositories();
```
mknoedel, line 115: It looks like we are calling out to fetch the repos multiple times. We should eventually end up using the JobState for this.

src/client.ts
```diff
@@ -97,65 +65,116 @@ export class APIClient {
...
+        token: string;
+        permissions: TokenPermissions;
+      };
+
+      //checking for proper scopes
+      if (
+        !(permissions.members === 'read' || permissions.members === 'write')
+      ) {
+        throw new IntegrationValidationError(
+          'Integration requires read access to organization members. See GitHub App permissions.',
```
mknoedel, line 137: Nice!

src/steps/account.ts
```diff
@@ -15,7 +14,9 @@ export async function fetchAccountDetails({
   instance,
   logger,
 }: IntegrationStepExecutionContext<IntegrationConfig>) {
-  const apiClient = createAPIClient(instance.config, logger);
+  const config = instance.config;
+  await sanitizeConfig(config);
```
mknoedel, line 18: I typically like to avoid mutation functions when possible, but that's more of a style thing.

src/sync/converters.ts
```diff
@@ -169,13 +172,17 @@ export function toPullRequestEntity(
...
     displayName: `${data.base.repo.name}/${data.number}`,
 
     accountLogin: data.base.repo.owner.login,
     repository: data.base.repo.name,
-    id: JSON.stringify(data.number),
+    //the type is hacked here because typing of data properties is controlled by a library call
+    //so I can't just say that data.number is a string
+    //here would be another way to solve it:
+    //id: JSON.stringify(data.number).replace(/\"/g, ''),
+    id: <string>(<unknown>data.number),
```
mknoedel, line 185: Isn't this a change from the old code. It used to be a string and now it is a number... maybe

src/sync/converters.ts
```diff
@@ -289,19 +297,21 @@ export function toTeamHasMemberRelationship(
...
   teamMemberEntity: UserEntity,
+  teamEntity: TeamEntity,
 ): TeamMemberRelationship {
   return {
     _key: `${teamMemberEntity._key}|manages|${teamEntity._key}`,
     _type: 'github_user_manages_team',
     _class: 'MANAGES',
     _fromEntityKey: teamMemberEntity._key,
     _toEntityKey: teamEntity._key,
+    displayName: 'MANAGES',
```
mknoedel, line 314: We eventually will want switch over to `createDirectRelationship` method form the sdk that sets things like the display name automatically.

src/util/createGitHubAppClient.ts
```diff
@@ -2,20 +2,28 @@ import { Octokit } from '@octokit/rest';
...
 import { IntegrationLogger } from '@jupiterone/integration-sdk-core';
+import { IntegrationConfig } from '../config';
 
-import fetchPrivateKey from '../util/fetchPrivateKey';
-
+//MAX_RETRIES is an arbitrary number, but consider that it limits the number
+// of retries for the entire client instance, not for a single request.
 const MAX_RETRIES = 20;
 
 export default async function createGitHubAppClient(
```
mknoedel, line 11: Looks like you will need to remove the await here, thanks GitHub!

src/util/fetchPrivateKey.ts
```diff
@@ -26,13 +28,13 @@ function getLocalPrivateKey(envLocalPathParam: string): string | undefined {
 async function getSsmPrivateKey(envSssParam: string) {
```
mknoedel, line 28: same here, remove the async

test/recording.ts
```diff
@@ -17,17 +17,16 @@ export function setupGithubRecording(
     mutateEntry: (entry) => {
       redact(entry);
     },
+    options: {
+      recordFailedRequests: true,
```
mknoedel, line 21: Making a note that we should eventually remove this once everything gets stable.

src/steps/teams.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    await jobState.addRelationship(
+      toOrganizationHasTeamRelationship(accountEntity, teamEntity),
+    );
+
+    for (const member of team.members || []) {
+      let memberEntity = (await jobState.findEntity(member.id)) as UserEntity;
+
+      if (!memberEntity) {
+        memberEntity = (await jobState.addEntity(
```
mknoedel, line 49: You can use a generic I believe in jobstate like `jobState.addEntity<UserEntity>`,
either that or just create the entity in a variable prior to adding it to the jobState

src/steps/teams.ts
```diff
@@ -0,0 +1,123 @@
...
+
+      if (member.role === TeamMemberRole.Maintainer) {
+        await jobState.addRelationship(
+          toMemberManagesTeamRelationship(memberEntity, teamEntity),
+        );
+      }
+    }
+
+    for (const repo of team.repos || []) {
+      let repoEntity = (await jobState.findEntity(repo.id)) as RepoEntity;
```
mknoedel, line 70: You can definitely do the generic with find entity - `jobState.findEntity<RepoEntity>(repo.id)`

src/steps/teams.ts
```diff
@@ -0,0 +1,123 @@
...
+        targetType: 'github_team',
+      },
+      {
+        _type: 'github_team_allows_repo',
+        _class: RelationshipClass.ALLOWS,
+        sourceType: 'github_team',
+        targetType: 'github_repo',
+      },
+    ],
+    dependsOn: ['fetch-repos', 'fetch-users'],
```
mknoedel, line 120: Eventually, it would be nice to pull these out into a constants file.

src/config.ts
```diff
@@ -70,17 +76,28 @@ export async function validateInvocation(
...
-    privateKeyEnvSsmParam: 'GITHUB_APP_PRIVATE_KEY', //this a hack. We need to define the real param.
-  });
+export function sanitizeConfig(config: IntegrationConfig) {
+  const localPath = process.env['GITHUB_APP_LOCAL_PRIVATE_KEY_PATH'];
+  if (localPath) {
+    let content;
+    try {
+      content = fs.readFileSync(localPath);
+    } catch (err) {
+      // basically not there
```
mknoedel, line 91: We should throw an error here as it is going to error at content.toString anyway

src/steps/account.ts
```diff
@@ -15,7 +15,6 @@ export async function fetchAccountDetails({
   logger,
 }: IntegrationStepExecutionContext<IntegrationConfig>) {
   const config = instance.config;
-  await sanitizeConfig(config);
```
mknoedel: Do we no longer need to get the config here?

kevincasey1222: Correct.

test/recording.ts
```diff
@@ -17,20 +17,13 @@ export function setupGithubRecording(
...
-      matchRequestsBy: {
-        headers: false,
-        order: false,
-      },
-    },
   });
 }
 
 function redact(entry): void {
-  if (entry.request.postData) {
```
mknoedel, line 32: Is there no sensitive postData anymore?

kevincasey1222: Right. I had that redacted before because in Auth0, the POST included the token, but here, it does not - the token is only in the Authorization header, which is already redacted. Also, the POST info is totally necessary in GraphQL recordings to see what the heck we requested.

mknoedel (approved):

PR #8, opened 2021-06-07
Typing lessons

kevincasey1222:
Hi @mknoedel !

Well, things are getting closer, but I might need some help cracking this final detail on recordings... 

As before, the tests all pass on local .env variables. In addition, the linting is fixed, the types are fixed, and we are all updated to the latest versions of Octokit (that took some sleuthing).

But, we're still failing to access the recordings under yarn:ci for many of the old tests, and I haven't figured out why. Maybe we could take a look at it together.

Kevin

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,19 +40,19 @@
     "uuid": "^3.0.0"
   },
   "peerDependencies": {
-    "@jupiterone/integration-sdk-core": "^6.3.0",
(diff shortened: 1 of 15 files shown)


kevincasey1222:
Per our Slack conversation, this PR is now dead - I'm pushing a new one called `cleanup` that doesn't have any secret leaks in the git hitory.

PR #9, opened 2021-06-08
Rebuild of previous pr typingLessons for gitleaks

kevincasey1222:
Hi @mknoedel !

So here is a cleanup rebuild of the previous PR that doesn't have any fake secrets in the git history.

The problem that is keeping it from passing the tests is something to do with authentication and the Polly recordings. The tests all pass with a local var.

The good news is that we're on the latest version of the @octokit modules now. 

--Kevin

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -119,62 +119,83 @@ export class APIClient {
   }
 
   public async setupAccountClient(): Promise<void> {
+    const authen: boolean = !(
+      this.config.githubAppPrivateKey === 'donotauthenticate'
+    ); //a hook for testing to skip certain authens if needed, because Polly isn't capturing all the authen
+
+    const installationId = Number(this.config.installationId);
+    const appClient = createGitHubAppClient(this.config, this.logger);
+    let myToken: string = '[REDACTED]';
+    let myPermissions: TokenPermissions = {
+      members: 'read',
+      metadata: 'read',
+    };
+
     try {
-      const installationId = Number(this.config.installationId);
-      const appClient = await createGitHubAppClient(this.config, this.logger);
-      const { token, permissions } = (await appClient.auth({
-        type: 'installation',
-      })) as {
-        token: string;
-        permissions: TokenPermissions;
-      };
-
-      //checking for proper scopes
-      if (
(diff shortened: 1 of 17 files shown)


src/client.ts
```diff
@@ -119,62 +119,83 @@ export class APIClient {
   }
 
   public async setupAccountClient(): Promise<void> {
+    const authen: boolean = !(
```
mknoedel, line 122: I think something like `shouldAuthenticate` would be a bit more clear.

src/client.ts
```diff
@@ -119,62 +119,83 @@ export class APIClient {
...
+      this.config.githubAppPrivateKey === 'donotauthenticate'
+    ); //a hook for testing to skip certain authens if needed, because Polly isn't capturing all the authen
+
+    const installationId = Number(this.config.installationId);
```
mknoedel, line 126: Probably want to validate with isNaN here

src/client.ts
```diff
@@ -119,62 +119,83 @@ export class APIClient {
...
+    const appClient = createGitHubAppClient(this.config, this.logger);
+    let myToken: string = '[REDACTED]';
```
mknoedel, line 128: This initialization pattern is a bit confusing. Could this be written in such a way where it is more obvious that the test environment has a token of '[REDACTED]' and so on?

mknoedel: As this is a bandaid anyway, perhaps an `isTest` checks an environment variable might suffice for now until we cut over.

kevincasey1222: Band aid removed

src/client.ts
kevincasey1222: Done

src/client.ts
kevincasey1222: Remvoed

src/util/sha.ts
```diff
@@ -5,3 +5,13 @@ export default function sha(text: string): string {
   sha256.update(text);
   return sha256.digest('hex').toString();
 }
+
+export function getFakeRsaKey(): string {
```
mknoedel, line 9: love it!

mknoedel (approved):

PR #11, opened 2021-06-09
Pull requests

kevincasey1222:
Hi @mknoedel !

This PR adds pull request entities to the project and is the first possibly complete version of the integration rewrite.

There are also a number of improvements per your feedback on previous PRs, plus cleaning out old code. There will still be another PR with further cleanups to the GraphQL client and other refactors that you've mentioned earlier. But I feel like enough has been done so far here that it's a good place for a review.

Thanks!
Kevin


src/steps/pullrequests.ts
```diff
@@ -0,0 +1,151 @@
...
+} from '../constants';
+
+export async function fetchPrs({
+  instance,
+  logger,
+  jobState,
+}: IntegrationStepExecutionContext<IntegrationConfig>) {
+  const config = instance.config;
+  const apiClient = createAPIClient(config, logger);
+  const accountEntity = (await jobState.getData(
```
mknoedel, line 41: I believe this could be `jobState.getData< AccountEntity >(DATA_ACCOUNT_ENTITY)`

kevincasey1222: Right you are. Updated.

src/steps/repos.ts
```diff
@@ -25,15 +31,21 @@ export async function fetchRepos({
...
     )) as RepoEntity;
 
+    repoEntities.push(repoEntity);
+
     await jobState.addRelationship(
       toAccountOwnsRepoRelationship(accountEntity, repoEntity),
     );
   });
+
+  await jobState.setData('REPO_ARRAY', repoEntities);
```
mknoedel, line 55: Is there a reason why we couldn't use `jobState.iterateEntities` here instead?

src/sync/converters.ts
```diff
@@ -271,11 +282,11 @@ export function toMemberManagesOrganizationRelationship(
 ): OrganizationMemberRelationship {
   return {
```
mknoedel, line 283: For these relationships, it would be nice to use the `createDirectRelationship` method from the new sdk, as these relationship methods don't really gain us much

src/sync/converters.ts
```diff
@@ -192,8 +203,8 @@ export function toPullRequestEntity(
     );
   }
   const entity: PullRequestEntity = {
```
mknoedel, line 205: Not important, but if you have time, it would be nice to pull each of these converter functions to converters unique to their step in order to make this file easier to read later. Not super important so ignore if you want.

mknoedel (approved):

PR #13, opened 2021-06-12
Final touches

kevincasey1222:
Hello @mknoedel !

Lots of cleanup in this PR. Cut 250 lines of vestigial code, put in support for "long account IDs use REST" thing, switched from .findEntity to passing arrays in jobState, went to createDirectRelationship, made PR entities "partial", and added function documentation.

Still to do in another PR: 
-- snapshots for pull requests
-- happy path tests for OrganizationAccountClient (including specific test for REST fallback on GraphQL error)
--more useful errors for util/createGithubAppClient retries
--double-check that checked scopes are sufficient
--GraphQL client code clean-up
--some kind of solution for synchronous PR ingestion

And that should be all. This rewrite is getting close to completion.

--Kevin


src/client/OrganizationAccountClient.ts
```diff
@@ -75,6 +75,22 @@ export default class OrganizationAccountClient {
     logger: IntegrationLogger;
     /**
      * Whether or not pull request commits should be analyzed for approval.
+     *
+     * Specifically, if this boolean is true, the pull-request entity gets
+     * for the following fields:
```
mknoedel, line 80: super unimportant, but we have been using `properties` instead of `fields` for entity keys.

kevincasey1222: Updated!

src/steps/index.test.ts
```diff
@@ -136,4 +138,26 @@ test('should collect data', async () => {
       required: ['webLink', 'displayName'],
     },
   });
+
```
mknoedel, line 141: Was this intentionally commented out?

src/sync/converters.ts
```diff
@@ -81,7 +63,7 @@ export function toTeamEntity(data: OrgTeamQueryResponse): TeamEntity {
     _type: GITHUB_TEAM_ENTITY_TYPE,
     _key: data.id,
     webLink: data.url,
-    name: data.slug,
+    name: data.slug, //this works, but why? Where is .slug set?
```
mknoedel, line 66: Looks like it returned in the graphQl response I think - https://docs.github.com/en/graphql/reference/objects#team

mknoedel (approved):
Great work well done!

src/steps/index.test.ts
kevincasey1222: Ah, no! Good catch. Leftover from dev.

kevincasey1222: Fixed.

PR #19, opened 2021-07-12
Add useRestForTeamRepos to more safely complete sdk 2 transition

mknoedel:
This should be removed directly after the initial transition as this is not a good pattern.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ and this project adheres to
 
 ## 1.0.0 - 2021-07-09
 
+### Added
+
+- new optional config variable `useRestForTeamRepos` that can is sometimes
+  needed to get around a GitHub error when fetching team repos via GraphQL.
+
 ### Changed
 
 - octokit packages to be dependencies instead of peer dependencies.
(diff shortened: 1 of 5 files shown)


kevincasey1222:
Hi @mknoedel ! I feel like this PR accomplishes the same goal as the try-catch starting on line 215 of https://github.com/JupiterOne/graph-github/blob/sdk-v2-release/src/client/OrganizationAccountClient.ts
My thought on that way was to try GraphQL, and if it fails, fall back to REST. If we make it a config option, then we have to predict which accounts will fail, right? Or are you planning to do both things - have the config option, and also have the try-catch as a backup plan?

mknoedel:
> Hi @mknoedel ! I feel like this PR accomplishes the same goal as the try-catch starting on line 215 of https://github.com/JupiterOne/graph-github/blob/sdk-v2-release/src/client/OrganizationAccountClient.ts
> My thought on that way was to try GraphQL, and if it fails, fall back to REST. If we make it a config option, then we have to predict which accounts will fail, right? Or are you planning to do both things - have the config option, and also have the try-catch as a backup plan?

It serves the same purpose except this is slightly safer for the initial cutover. We'll probably remove this after this gets released.

aiwilliams (approved):

PR #25, opened 2021-07-25
INT-1274 - Added merge properties on `github_pull_request`

austinkelleher:
- `mergedAt`
- `mergeCommitHash`


mknoedel (dismissed):

src/steps/__snapshots__/index.test.ts.snap
```diff
@@ -384,7 +384,9 @@ Object {
       "description": "Here are some things which are less important, but still things, nonetheless....",
       "displayName": "Test-repo/2",
       "id": 2,
+      "mergeCommitHash": "0ef05802b6094b61bf16aee65c0d157f407c02b0",
       "merged": false,
+      "mergedAt": undefined,
```
erkangz, line 389: we usually have `...On` as the time properties. Should we be consistent?

austinkelleher: I agree. This was an oversight 😅. Fixed! Thanks.

mknoedel (approved):

PR #27, opened 2021-08-10
Ingest collaborators

kevincasey1222:
Hi @aiwilliams ! Here is code to ingest GitHub collaborators, using the REST API. There is also a little stub code in here to do it with GraphQL for whenever we want to try to pick that up. The API call impact is one call per repo. Memory impact is three more sizable objects saved in the jobstate - two arrays and a map. (in a different architecture, I could have two API calls per repo and only two additional objects in the jobstate - I figured less API calls was the priority).

There's a fair bit going on here in order to accurately cover all scenarios (users that are added to a repo via multiple teams and/or directly, with different possible permissions, plus outside collaborator users). I think the architecture is sound - you may have some feedback on more efficient syntax.

I can add some more tests and recordings for outside collaborators once we are sure that this approach is right.

Also adding @ceelias and @mknoedel for possible reviews. A few eyes on this wouldn't be bad. If anyone wants a walkthrough of what is happening and why, I'm happy to oblige.

--Kevin


docs/jupiterone.md
```diff
@@ -111,12 +111,13 @@ The following relationships are created/mapped:
...
 | `github_account`      | **HAS**               | `github_user`         |
 | `github_account`      | **OWNS**              | `github_repo`         |
+| `github_repo`         | **ALLOWS**            | `github_team`         |
+| `github_repo`         | **ALLOWS**            | `github_user`         |
 | `github_repo`         | **HAS**               | `github_pullrequest`  |
-| `github_team`         | **ALLOWS**            | `github_repo`         |
 | `github_team`         | **HAS**               | `github_user`         |
 | `github_user`         | **APPROVED**          | `github_pullrequest`  |
 | `github_user`         | **MANAGES**           | `github_account`      |
-| `github_team`         | **MANAGES**           | `github_team`         |
```
mknoedel, line 119: Was this a mistake that a team managed a team?

src/client.ts
```diff
@@ -155,6 +157,37 @@ export class APIClient {
...
+    }
+    const collaborators: any = await this.accountClient.getRepoCollaboratorsWithRest(
+      repo.name,
+    );
+    for (const collab of collaborators) {
+      await iteratee(collab);
+    }
+
+    //let's see if GraphQL can get all collabs
+    //this is how we would prefer to do it, but it's not working right now
```
mknoedel, line 180: Is this not working on Github's side our our side?

src/client/GraphQLClient/resourceMetadataMap.ts
```diff
@@ -113,7 +113,30 @@ export default function(
...
+          }
+         
+        },
+        nodes {
+          ...userFields
+        }
+        ${pageInfo}
+      }`,
+      parent: OrganizationResource.Repositories,
+    },
```
mknoedel, line 140: It might be worth taking some more time trying to figure out if we could get this to work, and if it doesn't, add a comment saying the reasons why we are using REST for this endpoint.

src/client/GraphQLClient/types/client.ts
```diff
@@ -36,6 +37,19 @@ interface Actor {
 
 export type OrgQueryResponse = Node & Actor;
 
+export interface OrgCollaboratorQueryResponse extends Node, Actor {
+  permissions: CollaboratorPermissions;
+  node_id: string; //this matches a member's id, whereas the id here is unique to the collab
```
mknoedel, line 42: What is the `this` and `here` being referenced in this comment?

src/client/OrganizationAccountClient.ts
```diff
@@ -298,6 +300,48 @@ export default class OrganizationAccountClient {
     return this.teamRepositories || [];
   }
 
+  async getRepoCollaboratorsWithRest(repoName: string): Promise<any> {
+    try {
+      const repoCollaborators = await this.v3.paginate(
+        'GET /repos/{owner}/{repo}/collaborators' as any, // https://docs.github.com/en/rest/reference/repos#list-repository-collaborators
```
mknoedel, line 306: Is the `as any` necessary? I think octokit should pull types from the parameter.

src/client/OrganizationAccountClient.ts
```diff
@@ -298,6 +300,48 @@ export default class OrganizationAccountClient {
...
+        ]);
+
+        this.collaborators = collaborators;
+
+        return rateLimitConsumed;
+      });
+    }
+
+    return this.collaborators || [];
+  }
```
mknoedel, line 343: Looks like this should be commented out as well as it isn't getting used atm.

src/steps/index.test.ts
```diff
@@ -43,6 +46,7 @@ test('should collect data', async () => {
   await fetchMembers(context);
   await fetchRepos(context);
   await fetchTeams(context);
+  await fetchCollaborators(context);
```
mknoedel, line 49: It would be nice to verify somehow that this method was creating at least one External Collaborator,  GITHUB_REPO_USER_RELATIONSHIP and GITHUB_REPO_USER_RELATIONSHIP
```
 const repoUserRelationships = context.jobState.collectedEntities.filter((e) =>
    e._type.includes('GITHUB_REPO_USER_RELATIONSHIP'),
  );
expect(repoUserRelationships.length).toBeGreaterThan(0)
... then anything else you want to verify about them
```

src/steps/pullrequests.ts
```diff
@@ -50,27 +50,38 @@ export async function fetchPrs({
...
   }
-  const memberEntities = await jobState.getData<UserEntity[]>('MEMBER_ARRAY');
-  if (!memberEntities) {
-    throw new IntegrationMissingKeyError(
-      `Expected to find memberEntities in jobState.`,
-    );
+  let userEntities = await jobState.getData<UserEntity[]>('USER_ARRAY');
+  if (!userEntities) {
+    //try the members-only (no outside collaborators) array
+    userEntities = await jobState.getData<UserEntity[]>('MEMBER_ARRAY');
```
mknoedel, line 56: The difference between 'USER_ARRAY' and 'MEMBER_ARRAY' is a bit confusing to me. Could they be named like `EXTERNAL_AND_MEMBER_USER_ARRAY` and `MEMBER_USER_ARRAY`, or something that denotes the differences a bit better?

src/steps/pullrequests.ts
```diff
@@ -50,27 +50,38 @@ export async function fetchPrs({
...
+  let userByLoginMap = await jobState.getData<IdEntityMap<UserEntity>>(
+    'USER_BY_LOGIN_MAP',
   );
-  if (!memberByLoginMap) {
-    throw new IntegrationMissingKeyError(
-      `Expected to find memberByLoginMap in jobState.`,
+  if (!userByLoginMap) {
+    //try the members-only (no outside collaborators) map
+    userByLoginMap = await jobState.getData<IdEntityMap<UserEntity>>(
+      'MEMBER_BY_LOGIN_MAP',
```
mknoedel, line 70: Same here with the more clear naming.

mknoedel: It might be nice to pull these strings out into a `constants.ts` folder as well which can give a bit more context in the variable name without needing to extend the value.

src/steps/pullrequests.ts
```diff
@@ -171,7 +182,7 @@ export const prSteps: IntegrationStep<IntegrationConfig>[] = [
         partial: true,
       },
     ],
-    dependsOn: ['fetch-repos', 'fetch-users'],
+    dependsOn: ['fetch-repos', 'fetch-users', 'fetch-collaborators'],
```
mknoedel, line 185: Same here, I think it is probably time to create a `constants.ts` file and store these variables there.

src/sync/converters.ts
```diff
@@ -253,3 +282,41 @@ export function toPullRequestEntity(
...
+    _class: RelationshipClass.ALLOWS,
+    _type: GITHUB_REPO_USER_RELATIONSHIP_TYPE,
+    _fromEntityKey: repo._key,
+    _toEntityKey: user._key,
+    displayName: RelationshipClass.ALLOWS,
+    admin: permissions.admin,
+    maintain: permissions.maintain,
+    push: permissions.push,
+    triage: permissions.triage,
+    pull: permissions.pull,
```
mknoedel, line 319: could this be `'adminPermission': permissions.admin` instead? I'm worried that it won't be that clear to people that "admin" would me admin access (though I could be wrong about that). Same with the other permissions as well,

src/steps/collaborators.ts
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    //needed to calculate best permissions for a user on many teams
+    const loginPermissionsMap: IdEntityMap<CollaboratorPermissions> = {};
+    const loginsForThisRepo: string[] = [];
+
+    const relsToTeams = repoTeamRelationships.filter(
+      (rel) => rel._fromEntityKey === repo._key,
+    );
+    for (const teamRel of relsToTeams) {
```
mknoedel, line 137: Can we pull out pieces of this logic and make unit tests for them? It would be nice for the future to have something that ensures that this logic remains consistent as this code changes in the future.

I also think pulling out some of this logic will be helpful for readability.

src/steps/collaborators.ts
```diff
@@ -0,0 +1,238 @@
...
+          collabArray.push(collabEntity);
+        }
+        const repoUserRelationship = createRepoAllowsUserRelationship(
+          repo,
+          collabEntity,
+          'outside',
+          collab.permissions,
+        );
+        await jobState.addRelationship(repoUserRelationship);
+      }
```
mknoedel, line 103: It would be nice if this was it's own step to keep things a bit more atomic.

mknoedel (commented):
This is great work, thank you! The main change I would like i breaking up the large fetchCollaborators function into smaller tested pieces. All other comments are small suggestions.

docs/jupiterone.md
kevincasey1222: Correct.

src/client/GraphQLClient/types/client.ts
kevincasey1222: Updated the comment for clarity.

src/client/OrganizationAccountClient.ts
kevincasey1222: True. This was leftover from the dev process.

kevincasey1222: Removing it.

src/client/OrganizationAccountClient.ts
kevincasey1222: Agree. Done. I'll leave it as a comment to jumpstart research on getting the graphQL version working.

src/sync/converters.ts
kevincasey1222: I like it. Done.

src/steps/index.test.ts
kevincasey1222: Cool thought. I added that.

src/steps/collaborators.ts
kevincasey1222: Better yet, after some new insights into how GitHub is operating, I was able to cut this logic out entirely.

src/steps/collaborators.ts
kevincasey1222: I was able to make a big simplification - this is now the only collaborator step.

src/steps/pullrequests.ts
kevincasey1222: Done.

src/steps/pullrequests.ts
kevincasey1222: Agree and done.

src/client.ts
kevincasey1222: I'm not sure. We're getting an empty object response. I presume my GraphQL query is malformed, but I haven't been able to figure it out.

src/client/GraphQLClient/resourceMetadataMap.ts
kevincasey1222: It's worth working on, yes, because it might reduce API calls. But I'm not sure how to move forward right now - I need more GraphQL research. I'm thinking to tackle this in another PR, along with some other efficiency improvements in PR handling.

kevincasey1222:
Coming back to this after a break, I was able to simplify and improve quite a bit. I feel comfortable merging it now, if you do.

mknoedel (approved):
💯

PR #29, opened 2021-08-16
Ingest apps

kevincasey1222:
Hi @aiwilliams , @mknoedel , @ceelias !

This PR supports the ingestion of GitHub Apps (INT-400). It is built on PR 27, so it might be easier to review and merge that one first, to simplify the review here. Or, we could close PR 27 and just do it all here? As you like.

Kevin


src/client/OrganizationAccountClient.ts
```diff
@@ -317,6 +362,33 @@ export default class OrganizationAccountClient {
...
+        org: 'octokit',
+        type: 'private',
+      });
+      if (reply.data.installations) {
+        return reply.data.installations;
+      }
+      this.logger.warn({}, 'Found no installed GitHub apps');
+      return [];
+    } catch (err) {
+      this.logger.warn(
```
mknoedel, line 383: Should this not be an IntegrationError?

src/client/OrganizationAccountClient.ts
```diff
@@ -317,6 +362,33 @@ export default class OrganizationAccountClient {
...
+        {},
+        'Error while attempting to ingest to installed GitHub apps',
+      );
+      return [];
+      //throw new IntegrationError(err);
```
mknoedel, line 388: What is the reason for this commented out error.

src/sync/converters.ts
```diff
@@ -59,13 +74,42 @@ export function toAccountEntity(data: OrgQueryResponse): AccountEntity {
   return accountEntity;
 }
 
+export function toAppEntity(data: OrgAppQueryResponse): AppEntity {
+  const appEntity: AppEntity = {
+    _class: [GITHUB_APP_ENTITY_CLASS],
+    _type: GITHUB_APP_ENTITY_TYPE,
+    _key: 'GitHubAppInstallation_' + data.id,
```
mknoedel, line 81: It would be nice to have this be a function like `getAppEntityKey` that contains this prefix. Having floating keys can be tricky.

src/types/entities.ts
```diff
@@ -8,6 +8,28 @@ export interface AccountEntity extends Entity {
...
+  createdAt: string;
+  updatedAt: string;
+  events: string[];
+  repositorySelection: string;
+  singleFileName: string;
+  hasMultipleSingleFiles: boolean;
+  singleFilePaths: string[];
+  suspendedBy: string;
+  suspendedAt: string;
+  //plus permissions.whatever fields drawn from the permissions API object
```
mknoedel, line 30: It would be nice to expand on this comment. Is this saying 'permissions.admin'...?

src/util/createGitHubAppClient.ts
```diff
@@ -58,6 +58,13 @@ export default function createGitHubAppClient(
         }
       },
     },
+    /*
+    log: {
+      debug: console.log,
+      info: console.log,
+      warn: console.warn,
+      error: console.error,
+    }, */
```
mknoedel, line 67: remove this comment

src/steps/apps.ts
```diff
@@ -0,0 +1,74 @@
...
+import { toAppEntity } from '../sync/converters';
+import { AccountEntity, AppEntity } from '../types';
+import {
+  GITHUB_ACCOUNT_ENTITY_TYPE,
+  GITHUB_APP_ENTITY_TYPE,
+  GITHUB_APP_ENTITY_CLASS,
+  GITHUB_ACCOUNT_APP_RELATIONSHIP_TYPE,
+} from '../constants';
+
+export async function fetchApps({
```
mknoedel, line 21: Can I get a test for this as well?

kevincasey1222:
Rebased to include big changes from ingestCollaborators PR

src/client/OrganizationAccountClient.ts
```diff
@@ -317,6 +362,33 @@ export default class OrganizationAccountClient {
     return this.members || [];
   }
 
+  async getInstalledApps(ghsToken): Promise<OrgAppQueryResponse[]> {
+    //unable to apply the ghsToken to the Octokit v3 REST client
+    //also unable to find the App node in the GraphQL hierarcy under Organization
+    //and so making a direct @octokit/request call here, since the number of installed Apps is limited
+    try {
+      const reply = await request(`GET /orgs/${this.login}/installations`, {
```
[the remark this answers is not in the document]
kevincasey1222: Correct, as far as I can tell. There is something weird happening here - when the v3 octokit instance `.auth` method is invoked, you get a token, but apparently REST calls to the v3 instance are using some other auth method behind the scenes, because we get 'bad credentials' when we call the App endpoint through the v3 instance. Yet, if I explicitly pass the token returned by the v3 instance to this `request` call, we get the data.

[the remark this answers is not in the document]
kevincasey1222: So, if I call `paginate`, that's part of the v3 instance, and the API call fails with a 'bad credentials' message, which GitHub tech support tells us is because we are using the wrong token. I've tried explicitly passing the new token as a header in the paginate call, and it does not help.

src/client/OrganizationAccountClient.ts
kevincasey1222: Yeah, you're right.

src/client/OrganizationAccountClient.ts
kevincasey1222: I was trying to decide whether an error should be thrown here, but clearly it should. Thanks for catching that.

src/sync/converters.ts
kevincasey1222: Cool. Done.

src/types/entities.ts
kevincasey1222: Done

src/util/createGitHubAppClient.ts
kevincasey1222: Okay.

src/client/OrganizationAccountClient.ts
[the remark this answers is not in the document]
kevincasey1222: I am totally open to other ideas here - I tried a lot of things and sank about 10 hours just into this little detail before I found anything that worked. But I agree that it's not ideal to have so many different ways of pulling data. I was thinking we can get away with not paginating here because we wouldn't expect hundreds of GitHub Apps installed in one org.

src/steps/apps.ts
kevincasey1222: It's a part of `index.test.ts`. Or did you mean something else?

src/client/OrganizationAccountClient.ts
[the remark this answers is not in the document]
mknoedel: Dang that sounds like it was quite the drag, thanks for looking into that! Can you update the comment above this call with your investigation? After that's done I think we are good to move forward.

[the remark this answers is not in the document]
kevincasey1222: Done!

mknoedel (approved):
Thank you for this great work!

PR #40, opened 2021-08-30
INT-1677 - Pull Request step step is too slow

mknoedel:
Added rolling ingestion to the  pull request step.

Also rewrote the `organizationClient` to take the whole integrationContext instead of the logger, jobState, config and execution history as it was starting to get hard to follow.

### DIFF
--- a/docs/jupiterone.md
+++ b/docs/jupiterone.md
@@ -91,7 +91,7 @@ NOTE: ALL OF THE FOLLOWING DOCUMENTATION IS GENERATED USING THE
 "j1-integration document" COMMAND. DO NOT EDIT BY HAND! PLEASE SEE THE DEVELOPER
 DOCUMENTATION FOR USAGE INFORMATION:
 
-https://github.com/JupiterOne/sdk/blob/master/docs/integrations/development.md
+https://github.com/JupiterOne/sdk/blob/main/docs/integrations/development.md
 ********************************************************************************
 -->
 
--- a/docs/jupiterone.md
+++ b/docs/jupiterone.md
@@ -91,7 +91,7 @@ NOTE: ALL OF THE FOLLOWING DOCUMENTATION IS GENERATED USING THE
 "j1-integration document" COMMAND. DO NOT EDIT BY HAND! PLEASE SEE THE DEVELOPER
 DOCUMENTATION FOR USAGE INFORMATION:
 
-https://github.com/JupiterOne/sdk/blob/main/docs/integrations/development.md
+https://github.com/JupiterOne/sdk/blob/master/docs/integrations/development.md
 ********************************************************************************
 -->
 
--- a/src/client.ts
+++ b/src/client.ts
@@ -2,6 +2,7 @@ import {
   IntegrationLogger,
   IntegrationValidationError,
   IntegrationProviderAuthenticationError,
+  IntegrationExecutionContext,
 } from '@jupiterone/integration-sdk-core';
 
 import { IntegrationConfig } from './config';
@@ -45,8 +46,7 @@ export class APIClient {
   ghsToken: string;
   scopedForApps: boolean;
   constructor(
-    readonly config: IntegrationConfig,
-    readonly logger: IntegrationLogger,
+    readonly context: IntegrationExecutionContext<IntegrationConfig>,
   ) {}
 
   public async verifyAuthentication(): Promise<void> {
@@ -96,8 +96,8 @@ export class APIClient {
 
     // Check 'useRestForTeamRepos' config variable as a hack to allow large github
     // accounts to bypass a Github error. Please delete this code once that error is fixed.
-    const allTeamRepos: OrgTeamRepoQueryResponse[] = this.config
-      .useRestForTeamRepos
+    const allTeamRepos: OrgTeamRepoQueryResponse[] = this.context.instance
+      .config.useRestForTeamRepos
       ? await this.accountClient.getTeamReposWithRest()
       : await this.accountClient.getTeamRepositories();
 
@@ -174,7 +174,9 @@ export class APIClient {
     );
     if (pullrequests) {
       for (const pr of pullrequests) {
-        await iteratee(pr);
+        if (pr) {
+          await iteratee(pr);
+        }
       }
     }
   }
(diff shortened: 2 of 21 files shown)


docs/jupiterone.md
```diff
@@ -91,7 +91,7 @@ NOTE: ALL OF THE FOLLOWING DOCUMENTATION IS GENERATED USING THE
 "j1-integration document" COMMAND. DO NOT EDIT BY HAND! PLEASE SEE THE DEVELOPER
 DOCUMENTATION FOR USAGE INFORMATION:
 
-https://github.com/JupiterOne/sdk/blob/main/docs/integrations/development.md
+https://github.com/JupiterOne/sdk/blob/master/docs/integrations/development.md
```
mknoedel, line 94: Why did this go to master? Do I need to update the sdk package?

- [ ] switch this back to main

src/client.ts
```diff
@@ -174,7 +174,9 @@ export class APIClient {
     );
     if (pullrequests) {
       for (const pr of pullrequests) {
-        await iteratee(pr);
+        if (pr) {
```
mknoedel, line 177: PRs can now be undefined if they're created date is before the last successful execution date.

src/client.ts
```diff
@@ -291,17 +297,10 @@ export class APIClient {
...
     });
   }
 }
-
-export function createAPIClient(
-  config: IntegrationConfig,
-  logger: IntegrationLogger,
-): APIClient {
-  return new APIClient(config, logger);
-}
```
mknoedel, line 307: Removed this method as it doesn't do anything

src/client/OrganizationAccountClient.test.ts
```diff
@@ -45,12 +53,14 @@ describe('handles undefined resources from graphql', () => {
...
-    logger: mockLogger,
+    context,
     analyzeCommitApproval: true,
   });
 
   test('getTeams', async () => {
-    await expect(client.getTeams()).resolves.toEqual([]);
+    const teamsResponse = [];
+    await expect(client.getTeams()).resolves.toEqual(teamsResponse);
+    expect(((client as any).teams = teamsResponse));
```
mknoedel, line 63: Needed to check that `client.teams` wasn't undefined as it would throw an error under some circumstances.

docs/jupiterone.md
ndowmon: Yep. The dependencies were recently updated. You probably just need to install the present dependencies.

src/client/OrganizationAccountClient.ts
```diff
@@ -480,15 +479,24 @@ export default class OrganizationAccountClient {
           owner: account.login,
           repo: repo.name,
           per_page: prCount,
+          sort: 'created',
+          direction: 'desc', // newest to oldest
```
mknoedel, line 483: These are the defaults but it seemed like it would be better to be explicit here.

src/client/OrganizationAccountClient.ts
```diff
@@ -480,15 +479,24 @@ export default class OrganizationAccountClient {
...
       this.v3RateLimitConsumed++;
 
       return pMap(
         pullRequests,
         async (pullRequest) => {
+          // Only ingest PR if it was not already ingested in the previous integration run
+          const createdOn = parseTimePropertyValue(pullRequest.created_at);
+          if (createdOn && createdOn < lastSuccessfulExecutionTime) {
+            return undefined;
+          }
```
mknoedel, line 499: There isn't a way to only query for PRs up to a specific day so it made sense to just always pull 100 and then check the `created_at` value, as it doesn't affect the rate limiting at all.

docs/jupiterone.md
mknoedel: Turns out you do just need to upgrade the sdk packages.

ndowmon:
> **mknoedel** 7 minutes ago Author Member
Turns out you do just need to upgrade the sdk packages.

@mknoedel FWIW You didn't need to bump the packages, just run 
```
rm -rf node_modules
yarn install --force
```

Since I think you just hadn't installed packages since they were bumped to `6.16.1`. Bumping to `6.18.0` is fine though

src/client/OrganizationAccountClient.ts
```diff
@@ -480,15 +479,24 @@ export default class OrganizationAccountClient {
           owner: account.login,
           repo: repo.name,
           per_page: prCount,
```
ndowmon, line 481: It's a shame that this v3 API does not allow us to filter on `created` or `updated` properties, since the UI does allow us to filter like `updated:>=2013-02-01`. Maybe this will be available in the future.

ndowmon (changes requested):

src/client/OrganizationAccountClient.test.ts
mknoedel: I think storing these entities in the client is an artifact from moving from sdk v1 to sdk v2. There doesn't seem to be any real value in doing this as far as I can tell, especially because we now have the job state. 

I'll take a look into how difficult it will be to removing `teams` an the other entity related properties from `OrganizationAccountClient`

mknoedel: Yeah, that is not an ideal pattern. Looking into why it was likely done this way, It looks like these entity storage properties like `teams` and `repositories` were put into the client because we attempt to get all (except for Pull Request) Github data in a single graphQL call `getAccount`, which retrieves
```
        this.members = members;
        this.teams = teams;
        this.teamMembers = teamMembers;
        this.repositories = repositories;
        this.account = organization![0];
```

Each individual step will then either just iterate what was already fetched from the `fetch-account` step, or, if it doesn't see any stored values, it will attempt to fetch those values again using graphQL. I think this may have been ported over incorrectly, so we might be double-fetching data right now. This should get fixed

- [ ] Make sure we are not double pulling data.

As for if it is important at all to pull a ton of data in a single API call in the `getAccount` step, we do need to be hyper cognizant of how many api calls we are making to Github as they have very aggressive rate limiting, so grabbing as much data as possible in a single API call is definitely a good thing. The entities that are being stored are fairly small in number so it will be unlikely that we will run into any memory concerns by storing these values (plus we have had it out in production storing these values and we haven't run into any memory issues yet).

image

I think the solution to this is two-fold:
1. Shove these previously client-stored values into the jobState.
2. Do not attempt to re-pull for data that was already fetched in `getAccount`
2.5. Name `getAccount` something more description of what it is actually doing.

src/client/OrganizationAccountClient.ts
mknoedel: Updated makes way more sense, I'll update that.

src/client/OrganizationAccountClient.test.ts
ndowmon: >Shove these previously client-stored values into the jobState.

Are you suggesting adding this raw array of `teams` to the job state, or adding the `github_team` entities, and referencing them in downstream steps as `await jobState.iterateEntities({ _type: 'github_team' }, async () => { ... })`?

mknoedel:
No longer needed

PR #41, opened 2021-08-31
Ingest org secrets

kevincasey1222:
Hello @ceelias , @aiwilliams !

Here are GitHub org and repo Secrets. There is more to do in another PR - I want to set the scope checking to disable to steps instead of using if statements inside the steps, and I need to add environmental secrets. But that version of the PR was getting too big, so I decided to get this part submitted now.

--Kevin


src/client.ts
```diff
@@ -131,6 +133,69 @@ export class APIClient {
     }
   }
 
+  /**
+   * Iterates each Github organization secret.
+   *
+   * @param iteratee receives each resource to produce entities/relationships
+   */
+  public async iterateOrgSecrets(
+    allRepos: RepoEntity[],
```
mknoedel, line 142: Some accounts have a bunch of repos (over 1k), so it will probably be good to limit this to some sort of identifier or to just the repo values that are needed to create the secret entity.

src/client/OrganizationAccountClient.ts
```diff
@@ -405,6 +407,81 @@ export default class OrganizationAccountClient {
...
+        'GET /orgs/{org}/actions/secrets/{secret_name}/repositories', //https://docs.github.com/en/rest/reference/actions#list-selected-repositories-for-an-organization-secret
+        {
+          org: this.login,
+          secret_name: secretName,
+          per_page: 100,
+        },
+        (response) => {
+          this.logger.info('Fetched page of repos for a secret');
+          this.v3RateLimitConsumed++;
+          return response.data;
```
mknoedel, line 450: can you turn all of these `get...` methods into `iterate...` methods? because there is a callback in `v3.paginate`, we should just add a `iteratee: ResourceIteratee<OrgSecretQueryResponse>` into the parameters and call it in the callback here.

In general, I have noticed that we are putting a lot into memory in this integration and I'm worrying that we are going to start running out of memory.

src/steps/orgsecrets.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  const accountEntity = await jobState.getData<AccountEntity>(
+    DATA_ACCOUNT_ENTITY,
+  );
+  if (!accountEntity) {
+    throw new IntegrationMissingKeyError(
+      `Expected to find Account entity in jobState.`,
+    );
+  }
+  const repoEntities = await jobState.getData<RepoEntity[]>(GITHUB_REPO_ARRAY);
```
mknoedel, line 41: Does this `GITHUB_REPO_ARRAY` need to contain the whole repo entity?

src/steps/orgsecrets.ts
```diff
@@ -0,0 +1,109 @@
...
+        await jobState.addRelationship(
+          createDirectRelationship({
+            _class: RelationshipClass.USES,
+            from: repoEntity,
+            to: secretEntity,
+          }),
+        );
+      }
+    }
+    orgSecretsByNameMap[secret.name] = secretEntity;
```
mknoedel, line 75: Is there ever going to be a name conflict? If so it might be nice to check first and log something.

src/util/propertyHelpers.ts
```diff
@@ -49,3 +49,13 @@ export function decomposePermissions(permissions: TokenPermissions) {
 export function getAppEntityKey(installId): string {
   return 'GitHubAppInstallation_' + installId;
 }
+
+export function getSecretEntityKey(
+  name,
+  secretOwnerType,
+  secretOwnerName,
+): string {
```
mknoedel, line 57: Super unimportant, but making the parameters be an object would be a bit more clear to me as the order would be confusing. Outside that, it might be worth adding a quick input validation.

src/steps/reposecrets.ts
```diff
@@ -0,0 +1,131 @@
...
+        }),
+      );
+      await jobState.addRelationship(
+        createDirectRelationship({
+          _class: RelationshipClass.USES,
+          from: repoEntity,
+          to: secretEntity,
+        }),
+      );
+      if (orgSecretEntities[secret.name]) {
```
mknoedel, line 79: Could there be multiple orgSecretEntities that override the same secret?

src/sync/converters.ts
```diff
@@ -103,6 +110,69 @@ export function toAppEntity(data: OrgAppQueryResponse): AppEntity {
   return appEntity;
 }
 
+export function toOrgSecretEntity(
+  data: OrgSecretQueryResponse,
+  orgLogin: string,
+): SecretEntity {
+  const secretEntity: SecretEntity = {
+    _class: [GITHUB_SECRET_ENTITY_CLASS],
```
mknoedel, line 118: I think it's a bit more clear to have `GITHUB_SECRET_ENTITY_CLASS` itself be an array so we can add to it explicitly in the constants file.

mknoedel (commented):
Cool new functionality! I added a few comments to address prior to merging.

kevincasey1222:
Rebased this off of Michael's adjusting to the GraphQL client and pull request systems (which were a big improvement). Now I can address the above comments.

src/client.ts
kevincasey1222: Okay, so I rewrote things to use `GITHUB_REPO_ARRAY` as an array of simpler objects of format `{ _key: string; name: string }`, which should reduce the memory impact substantially from passing around the array of repo entities.

src/steps/orgsecrets.ts
kevincasey1222: Changed this so it doesn't anymore

src/steps/orgsecrets.ts
kevincasey1222: At the org secret level, I don't think there can be a name conflict because the secrets are referenced by name, like constants.

src/steps/reposecrets.ts
kevincasey1222: I think org secret names have to be unique. Repo secret names are unique in the repo. So we could have multiple repos with a named secret that overrides the org secret of that name, but not multiple org secrets of the same name.

src/sync/converters.ts
kevincasey1222: Cool. Done.

src/util/propertyHelpers.ts
kevincasey1222: Cool. Did that.

src/steps/orgsecrets.ts
kevincasey1222: I did add a log warning just in case.

src/client/OrganizationAccountClient.ts
kevincasey1222: Fair enough. Could we do that change in a future PR?

src/client.ts
```diff
@@ -124,6 +131,69 @@ export class APIClient {
...
+  public async iterateOrgSecrets(
+    allRepos: RepoKeyAndName[],
+    iteratee: ResourceIteratee<OrgSecretQueryResponse>,
+  ): Promise<void> {
+    if (!this.accountClient) {
+      await this.setupAccountClient();
+    }
+    if (this.secretsScope) {
+      const secrets: OrgSecretQueryResponse[] = await this.accountClient.getOrganizationSecrets();
+      for (const secret of secrets) {
```
mknoedel, line 148: are we concerned that secrets will ever be `undefined`?

kevincasey1222: `getOrganizationSecrets` concludes currently with `return this.orgSecrets || [];`
Does that feel safe, or should we also be defensive here in case of future code changes?

src/steps/repos.ts
```diff
@@ -55,7 +55,7 @@ export async function fetchRepos({
     );
   });
 
-  await jobState.setData(GITHUB_REPO_ARRAY, repoEntities);
+  await jobState.setData(GITHUB_REPO_ARRAY, repoTags);
```
mknoedel, line 58: Could this be REPO_TAGS_ARRAY or REPO_KEY_AND_NAME_ARRAY?

mknoedel (approved):

PR #56, opened 2021-09-15
More robustness for one repo fail scenarios

kevincasey1222:
Hi @mknoedel !

Here you go!

--Kevin

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -224,13 +224,22 @@ export class APIClient {
     if (!this.accountClient) {
       await this.setupAccountClient();
     }
-    const {
-      rateLimitConsumed,
-    } = await this.accountClient.iteratePullRequestEntities(repo, iteratee);
-    logger.info(
-      { rateLimitConsumed },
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -245,11 +254,24 @@ export class APIClient {
     if (!this.accountClient) {
       await this.setupAccountClient();
     }
-    const collaborators: OrgCollaboratorQueryResponse[] = await this.accountClient.getRepoCollaboratorsWithRest(
-      repo.name,
-    );
-    for (const collab of collaborators) {
-      await iteratee(collab);
+    try {
+      const collaborators: OrgCollaboratorQueryResponse[] = await this.accountClient.getRepoCollaboratorsWithRest(
```
mknoedel, line 258: `getRepoCollaboratorsWithRest` already has a try catch in it so we likely just need to log better in that method than add a try catch here.

kevincasey1222: Okay, cool. Done.

mknoedel (approved):

PR #62, opened 2021-09-17
Extra props

kevincasey1222:
Hi @mknoedel !

Here are lots of extra properties per request from NextRoll, INT-1652.

--Kevin


src/sync/converters.ts
```diff
@@ -349,6 +399,35 @@ export function createRepoAllowsUserRelationship(
...
+    _mapping: {
+      sourceEntityKey: prKey,
+      relationshipDirection: RelationshipDirection.REVERSE,
+      targetFilterKeys: [['_type', 'login']],
+      targetEntity: {
+        _class: 'User',
+        _type: GITHUB_MEMBER_ENTITY_TYPE,
+        login: unknownLogin,
+      },
+      skipTargetCreation: true,
```
mknoedel, line 425: This needs to be false for the comment to be true.

src/sync/converters.ts
```diff
@@ -376,6 +455,19 @@ export function toPullRequestEntity(
...
   );
+  const commitsCount = commits ? commits.length : 0;
+  const approvalsCount = reviews ? reviews.filter(isApprovalReview).length : 0;
+
+  let approvalLastAt: number | undefined = undefined;
+  if (approvedCommits) {
+    const commitTimes = approvedCommits?.map(
+      (c) => parseTimePropertyValue(c.authoredDate) || 0,
+    );
+    const maxTime = Math.max(...commitTimes);
```
mknoedel, line 466: Does this throw an error with a spread empty array?

mknoedel (commented):
Looks great! A few small comments

src/sync/converters.ts
kevincasey1222: Good catch! Thanks!

src/sync/converters.ts
kevincasey1222: Also good catch! Thanks!

mknoedel (approved):

PR #80, opened 2021-10-18
JobLog event for 403 errors on environment secrets

mknoedel:


src/client/OrganizationAccountClient.ts
```diff
@@ -402,15 +402,20 @@ export default class OrganizationAccountClient {
...
+      if (err.status == '403') {
         // Don't fail step if integration does not have access to secrets.
-        throw new IntegrationError({
-          message: err.message,
-          code: err.status,
-          cause: err,
+        this.logger.publishEvent({
+          name: 'UNAUTHORIZED',
+          description:
+            'Unable to ingest environment secrets. Ensure scope secrets:read is enabled',
```
austinkelleher, line 410: Nice! Is it possible to disable this step entirely based on what scopes are actually enabled?

PR #89, opened 2021-10-21
Better graph ql errors

kevincasey1222:
Hi @austinkelleher !

Can you take a look at this and see if I'm doing it the right way? The goal is to catch [200] responses from GraphQL that are actually rate limiting errors. (INT-1952, linked to many Sentry errors).

--Kevin


src/client/GraphQLClient/client.ts
```diff
@@ -358,8 +361,28 @@ export class GitHubGraphQLClient {
...
+     * {"message":"API rate limit exceeded for 98.53.189.133. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
+     *
+     * Check https://github.com/lifeomic/attempt for options on retry
+     *
+     */
+
+    const queryWithRateLimitCatch = async () => {
+      const response = await query();
+      if (response.message?.includes('API rate limit exceeded')) {
+        throw new IntegrationError({
```
austinkelleher, line 377: Maybe we should update this to a `IntegrationProviderAPIError`. Thoughts?

src/client.ts
```diff
@@ -361,7 +361,13 @@ export class APIClient {
       });
     }
 
-    this.processScopes(myPermissions);
+    if (myPermissions) {
+      this.processScopes(myPermissions);
+    } else {
+      throw new IntegrationValidationError(
+        'Module setupAccountClient failed to receive any permissions from client authentication.',
```
austinkelleher, line 368: Is this an unrelated change? Is this something that we are seeing from Sentry? If it's user error, we should consider a more targeted error message that explains how to fix the issue/points to docs. Thoughts??

src/client/GraphQLClient/client.ts
kevincasey1222: Good though. I'll do it.

PR #95, opened 2021-11-02
Token refresh

kevincasey1222:
Hi @mknoedel !

Here is the code to allow the GraphQL client to update its token when it is about to expire.

I built this off the the Updated mfaEnabled prop to undefined when null PR, so this includes those changes too, and a few misc fixes I discovered along the way.


src/steps/reposecrets.ts
```diff
@@ -129,7 +129,7 @@ export const repoSecretSteps: IntegrationStep<IntegrationConfig>[] = [
         targetType: GITHUB_ORG_SECRET_ENTITY_TYPE,
       },
     ],
-    dependsOn: ['fetch-org-secrets'],
```
mknoedel, line 132: Do we still need this step to depend on `fetchhrepos`?

src/steps/envsecrets.ts
```diff
@@ -144,7 +144,11 @@ export const envSecretSteps: IntegrationStep<IntegrationConfig>[] = [
         targetType: GITHUB_REPO_SECRET_ENTITY_TYPE,
       },
     ],
-    dependsOn: ['fetch-environments', 'fetch-repo-secrets'],
+    dependsOn: [
+      'fetch-environments',
+      'fetch-repo-secrets',
+      'fetch-org-secrets',
```
mknoedel, line 150: good catch with this one, this might have been causing some problems as well.

mknoedel (approved):
Looks great thanks!

src/steps/reposecrets.ts
kevincasey1222: Yeah, still needs that, so populate repo tags.

PR #105, opened 2021-11-08
Improving structure and tests

kevincasey1222:
Hi @austinkelleher !

This PR moves some functions that we did last week into /util, adds better tests and typing, fixes some old tests that stopped being meaningful a few commits ago, and cleans up junk code and comments.


kevincasey1222:
I updated the GraphQL client package in hopes that it might fail more gracefully on GitHub API 503 errors, which are occasionally appearing in Sentry as a JSON parsing problem.

kevincasey1222:
I promise I'm done modifying this PR now. :) You can review it. I'll go work on something else.

src/client/GraphQLClient/client.ts
```diff
@@ -110,24 +112,44 @@ export class GitHubGraphQLClient {
...
+          pullRequestQueryString,
+          async () => {
+            return await queryPullRequests({
+              query,
+              ...queryCursors,
+            });
+          },
+        );
+      } catch (err) {
+        if (err.status === 401) {
```
austinkelleher, line 137: It seems like this logic for handling 401 status code is used in several places. Is there a reason why we wouldn't enhance `retryGraphQL` to handle this case?

austinkelleher (approved):

PR #107, opened 2021-11-15
Breaking up teams step

kevincasey1222:
Separating the fetch-teams step into 3 steps : fetch-teams, fetch-team-members, and fetch-team-repos

This will help isolate errors, because fetch-teams for just the entities is actually pretty quick compared to the other two.


aiwilliams (dismissed):
A couple of questions but LGTM. May be good to get a review from @mknoedel , whom I've requested it from.

aiwilliams:
Just learned that @austinkelleher has been helping on this integration, so requested review from him.

src/steps/teams.ts
```diff
@@ -57,51 +47,6 @@ export async function fetchTeams({
...
       }),
     );
-
-    for (const member of team.members || []) {
-      const memberKey = member.id;
-      if (!(await jobState.hasKey(memberKey))) {
-        await jobState.addEntity(
-          toOrganizationMemberEntityFromTeamMember(member),
-        );
-      }
```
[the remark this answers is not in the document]
kevincasey1222: Yep

src/client/OrganizationAccountClient.ts
```diff
@@ -441,8 +441,15 @@ export default class OrganizationAccountClient {
           return response?.data;
         },
       );
-      return repoSecrets || [];
+      return repoEnvSecrets || [];
     } catch (err) {
+      if (err.status === 403) {
+        //this is caused by repos with more restrictive privacy settings
+        this.logger.info(
+          `Repo ${repoName} returned a 403 unauthorized when environmental secrets requested.`,
```
austinkelleher, line 449: Can we change this to have `repoName` as an object property in the first arg? We should avoid string interpolation/concatenation of log messages when possible. Static log messages are easier and faster to search for. 

Searching for an exact match of `"Repo returned a 403 unauthorized when environmental secrets requested."` is faster than leveraging the "contains partial" operator and searching for something like `"returned a 403 unauthorized when environmental secrets requested."`

```suggestion
          { repoName },
          'Repo returned a 403 unauthorized when environmental secrets requested.',
```

src/client/OrganizationAccountClient.ts
```diff
@@ -428,7 +428,7 @@ export default class OrganizationAccountClient {
     repoName: string,
   ): Promise<SecretQueryResponse[]> {
     try {
-      const repoSecrets = await this.v3.paginate(
+      const repoEnvSecrets = await this.v3.paginate(
```
austinkelleher, line 431: Nice variable rename

src/steps/teamMembers.ts
```diff
@@ -0,0 +1,92 @@
...
+    throw new IntegrationMissingKeyError(
+      `Expected to find Account entity in jobState.`,
+    );
+  }
+
+  await apiClient.iterateTeamMembers(async (user) => {
+    if (!(await jobState.hasKey(user.id))) {
+      //somehow this team has a user we didn't know about
+      //shouldn't happen, except through weird timing, but we'll make an entry
+      await jobState.addEntity(toOrganizationMemberEntityFromTeamMember(user));
```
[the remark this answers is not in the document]
austinkelleher: I think the main implication is around partial datasets. If this step fails, the `github_user` `_type` will be marked as partial and the persister will not process deletes of `github_user`.

src/steps/teamRepos.ts
```diff
@@ -0,0 +1,62 @@
...
+  logger,
+  jobState,
+}: IntegrationStepExecutionContext<IntegrationConfig>) {
+  const config = instance.config;
+  const apiClient = createAPIClient(config, logger);
+
+  await apiClient.iterateTeamRepos(async (teamRepo) => {
+    if (!jobState.hasKey(teamRepo.id)) {
+      throw new IntegrationMissingKeyError(
+        `Expected repo (CodeRepo) with id to exist (key=${teamRepo.id})`,
```
austinkelleher, line 28: I realize this is old code, but now that I'm reviewing it...

I think that it's possible that a new repo between between the time that the repo step finished, and this step starts. The timing of that would cause this entire step to fail. I think we should just `warn` here. If it happens a lot, then it may be an indication that we have a wider issue, but I don't think we should fail the entire step in this case.

src/steps/teamRepos.ts
```diff
@@ -0,0 +1,62 @@
...
+export async function fetchTeamRepos({
+  instance,
+  logger,
+  jobState,
+}: IntegrationStepExecutionContext<IntegrationConfig>) {
+  const config = instance.config;
+  const apiClient = createAPIClient(config, logger);
+
+  await apiClient.iterateTeamRepos(async (teamRepo) => {
+    if (!jobState.hasKey(teamRepo.id)) {
```
austinkelleher, line 26: `hasKey` is `async`. Can we please `await` this?

src/steps/teamRepos.ts
```diff
@@ -0,0 +1,62 @@
...
+      throw new IntegrationMissingKeyError(
+        `Expected repo (CodeRepo) with id to exist (key=${teamRepo.id})`,
+      );
+    }
+    //property .teams is just a single key
+    if (!jobState.hasKey(teamRepo.teams)) {
```
austinkelleher, line 32: `hasKey` is `async`. Can we please `await` this?

src/steps/teamRepos.ts
```diff
@@ -0,0 +1,62 @@
...
+      throw new IntegrationMissingKeyError(
+        `Expected team (UserGroup) with id to exist (key=${teamRepo.teams})`,
```
austinkelleher, line 34: Same comment as above about not failing the entire step if we are missing the key.

austinkelleher (changes requested):
Thanks for improving this!

src/steps/teamMembers.ts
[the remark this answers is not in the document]
kevincasey1222: Good point. Thanks.

src/client/OrganizationAccountClient.ts
kevincasey1222: Great point - since I'm not directly using Cloudwatch, I didn't think of that.

src/steps/teamRepos.ts
kevincasey1222: Agree.

src/steps/teamRepos.ts
kevincasey1222: Yes! Thanks for catching that

src/steps/teamRepos.ts
kevincasey1222: Yes! Turns out, that mistake was in a few steps, including collaborators, which has been slow lately... hmmm... fixed everywhere.

src/steps/teamRepos.ts
kevincasey1222: Done.

kevincasey1222:
Updated those `await jobstate.hasKey` calls, improved logs, and made improvements per your comments above.

src/util/sleepIfApproachingRateLimit.ts
```diff
@@ -30,21 +30,22 @@ export default async function sleepIfApproachingRateLimit(
...
+          {
+            rateLimitPercentLeft: (
+              rateLimitRemainingProportion * 100
+            ).toPrecision(4),
+            secondsUntilReset: msUntilRateLimitReset / 1000,
+            resetAt: rateLimit.resetAt,
+          },
+          `Rate limit almost exhausted, sleeping until reset time`,
+        );
+        await sleep(msUntilRateLimitReset);
```
austinkelleher, line 47: Thank you!

austinkelleher (approved):

src/steps/teamMembers.ts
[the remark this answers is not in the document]
aiwilliams: Thanks, updated the development guide outline.

PR #112, opened 2021-11-20
Ints 70 cursors fix

kevincasey1222:
Hi guys,

The problem was that when we had nested objects that were large enough for GraphQL to need multiple API calls, the cursors were resetting and starting over.

The recursive cursor logic for nested objects is tough to analyze with absolute certainty, but I found the problem by changing the pagination limits in `queries.ts` to `1`.  This PR fixes the problem under those conditions and produces the correct snapshots for my test account. 

Naturally, for the real code, I had to change the query limits back, but I can say for certain that the changes in this PR are confined to what is already broken, so it won't make things any worse, and it does seem to solve the problem.

### DIFF
--- a/src/client/GraphQLClient/response.ts
+++ b/src/client/GraphQLClient/response.ts
@@ -6,7 +6,7 @@ import {
 } from './types';
 
 export function mapResponseCursorsForQuery(
-  cursors: ResourceMap<CursorHierarchy>,
+  pageCursors: ResourceMap<CursorHierarchy>,
   queryCursors: ResourceMap<string>,
 ): ResourceMap<string> {
   function cursorsFromHierarchy(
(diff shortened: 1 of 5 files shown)


src/client/GraphQLClient/response.ts
```diff
@@ -31,6 +31,10 @@ export function mapResponseCursorsForQuery(
 
       if (queryCursors[key]) {
         cursors[key] = queryCursors[key];
+      } else {
+        if (hierarchy.self) {
+          cursors[key] = hierarchy.self;
+        }
```
kevincasey1222, line 37: This seems to be the key. The rest of this PR is just linting and a variable rename.

austinkelleher: Would be great to pair this with tests. The logic is complex! 😅

austinkelleher: With that said, I see how this could potentially be a problem after stepping through logs a bit. 🤔  I think we should roll forward with it.

austinkelleher (approved):

PR #114, opened 2021-11-22
Stream processing repo collabs

kevincasey1222:
Hi @austinkelleher @aiwilliams ,

Just a draft PR for now so you can see the progress.

I've got the change made for stream processing repo collaborators and it works great. Wasn't as simple as I expected because of `extractSelectedResources` being complicated, but it works great now.

I thought this would make the changes to `fetch-team-repos` and `fetch-team-members` easy, but it turns out there are more details to work through because unlike Repos, Teams aren't top-level objects in GraphQL. But it's going to work. Just need a few more minutes to sort it out.

### DIFF
--- a/src/client/OrganizationAccountClient.ts
+++ b/src/client/OrganizationAccountClient.ts
@@ -45,6 +45,7 @@ import {
   TEAMS_QUERY_STRING,
   USERS_QUERY_STRING,
   COLLABORATORS_QUERY_STRING,
+  SINGLE_REPO_COLLABORATORS_QUERY_STRING,
 } from './GraphQLClient/queries';
 
 export default class OrganizationAccountClient {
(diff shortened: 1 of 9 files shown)


src/client/OrganizationAccountClient.ts
```diff
@@ -253,6 +254,28 @@ export default class OrganizationAccountClient {
...
+          },
+        );
+
+      if (collaborators) {
+        response = response.concat(collaborators as Collaborator[]);
+      }
+      return rateLimitConsumed;
+    });
+    return response;
+  }
```
aiwilliams, line 277: This means that we're collecting all collaborators into memory before iterating the collaborators in a single page, so in fact we're not stream processing the data. I think we need a new function that will allow us to call the iteratee function before moving to the next page.

kevincasey1222: True. But we're just collecting all collaborators for the one repo into memory, so it's not as many. It's "streamier". ;)

But yes, we need to pass the iterator down deeper, but I was thinking to make that a future PR.

PR #115, opened 2021-11-22
Streamier teams

kevincasey1222:
Hi @austinkelleher and @aiwilliams !

This one is much better - it has the previous PR 114's change that makes collaborators work properly, and it has an improvement to `fetch-team-repos` and `fetch-team-members`. They will now pull the first 100 relationships per team, and if there are more than that for a given team, those will be ignored and the step will proceed.

Obviously we still have to fix that limitation, but this is deployable and should let everyone's integrations complete properly and in a timely fashion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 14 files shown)


aiwilliams:
@kevincasey1222 I'm running this branch locally against our GitHub organization now, will report back.

PR #120, opened 2021-11-29
INTS-113 & INTS-114 - Only add relationships if we haven't seen the key before

mknoedel:


src/util/safeAddRelationship.ts
```diff
@@ -0,0 +1,14 @@
+import { JobState, Relationship } from '@jupiterone/integration-sdk-core';
+
+export async function safeAddRelationship(
+  jobState: JobState,
+  relationshipKeys: Set<string>,
+  relationship?: Relationship,
+) {
+  if (relationship && !relationshipKeys.has(relationship._key)) {
+    await jobState.addRelationship(relationship);
+    relationshipKeys.add(String(relationship._key));
```
austinkelleher, line 10: Do we have to use `String` here?

austinkelleher (commented):
Is the `_key` actually unique enough? Under what circumstance would there be a duplicate key?

src/util/safeAddRelationship.ts
mknoedel: We do not. I'll remove that.

mknoedel:
I'm going to do this in a better way.

PR #122, opened 2021-11-29
Fix Graphql children cursor retry logic

mknoedel:
There were 2 problems with the graphQL response cursing:
1. Only the base-level query cursor was getting read, which meant any child-level cursing was getting ignored.
2. Only the base-level resource was being used when determining there was a next page or not, which was preventing repos from getting ingested.

Passed against dev locally
image

### DIFF
--- a/src/client/GraphQLClient/client.ts
+++ b/src/client/GraphQLClient/client.ts
@@ -20,7 +20,8 @@ import {
   SINGLE_PULL_REQUEST_QUERY_STRING,
 } from './queries';
 import {
-  extractSelectedResources,
+  responseHasNextPage,
+  processGraphQlPageResult,
   mapResponseCursorsForQuery,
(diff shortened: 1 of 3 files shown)


kevincasey1222 (commented):
A few notes for reviewers: (I paired with Michael on this one):

Changes in `src/client/GraphQLClient/client.ts` are really just variable renaming and pulling the `responseHasNextPage` function out into a function where we could experiment.

Change in `src/client/GraphQLClient/response.test.ts` are renaming and clarifying the purposes of tests, and adding a few more for the `responseHasNextPage` function

The changes in behavior are in `response.ts`

austinkelleher (dismissed):

a-u-h-g (approved):

austinkelleher (approved):

PR #124, opened 2021-11-30
Int 2216 catch etimedout node error

kevincasey1222:
Hi @mknoedel or @austinkelleher !

Here is an update to handle the ETIMEDOUT node error and move the sleep function out of the retry logic.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -119,10 +119,17 @@ export class APIClient {
     if (!this.accountClient) {
       await this.setupAccountClient();
     }
-    const teamRepos: OrgTeamRepoQueryResponse[] =
-      await this.accountClient.getTeamRepositories(teamSlug);
-    for (const teamRepoAssociation of teamRepos) {
-      await iteratee(teamRepoAssociation);
+    try {
+      const teamRepos: OrgTeamRepoQueryResponse[] =
+        await this.accountClient.getTeamRepositories(teamSlug);
+      for (const teamRepoAssociation of teamRepos) {
+        await iteratee(teamRepoAssociation);
+      }
+    } catch (err) {
+      this.logger.warn(
+        { teamSlug, err },
+        '[SKIP] Problem while ingesting repos accessed by a team.',
+      );
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -119,10 +119,17 @@ export class APIClient {
...
+    try {
+      const teamRepos: OrgTeamRepoQueryResponse[] =
+        await this.accountClient.getTeamRepositories(teamSlug);
+      for (const teamRepoAssociation of teamRepos) {
+        await iteratee(teamRepoAssociation);
+      }
+    } catch (err) {
+      this.logger.warn(
+        { teamSlug, err },
+        '[SKIP] Problem while ingesting repos accessed by a team.',
```
mknoedel, line 131: I'm afraid that handling the error in this way will cause deletions of the previous run as when the diffing with the graph happens, it won't find the change.  We could mark the step as partial which prevents diffing for deletions, though I don't know how to do that off the top of my head.

src/client.ts
```diff
@@ -138,10 +145,17 @@ export class APIClient {
...
+      const teamMembers: OrgTeamMemberQueryResponse[] =
+        await this.accountClient.getTeamMembers(teamSlug);
+      for (const teamUserAssociation of teamMembers) {
+        await iteratee(teamUserAssociation);
+      }
+    } catch (err) {
+      this.logger.warn(
+        { teamSlug, err },
+        '[SKIP] Problem while ingesting members of a team.',
+      );
```
mknoedel, line 158: Same issue as above.

src/client.ts
kevincasey1222: Right. I worried about that too. I mean, we could fail the whole step. It's just likely that the error hits on these long steps.

src/client/GraphQLClient/client.ts
```diff
@@ -689,8 +692,20 @@ export class GitHubGraphQLClient {
...
+        );
+      },
+      handleTimeout(attemptContext: AttemptContext) {
+        logger.warn(
+          { attemptContext },
+          `Attempt to contact GraphQL timed out after 3 minutes.`,
         );
+        throw new IntegrationProviderAPIError({
+          status: 'None',
+          statusText: `GraphQL query timeout error: ${queryString}`,
```
mknoedel, line 705: Need to put something like TIMEOUT here.

kevincasey1222: I put it in the endpoint. Does that work?

mknoedel: Reading it again, this is quite clear. Idk what I was thinking 🙃

src/client/OrganizationAccountClient.ts
```diff
@@ -622,15 +622,28 @@ export default class OrganizationAccountClient {
...
-      const errors = responseErrors.errors
-        ? responseErrors.errors
-        : [responseErrors];
-      throw new IntegrationError({
-        message: name + ': ' + errors.map((e) => e.message).join(' | '),
-        code: errors[0].Code,
-        cause: errors[0].stack ? errors : JSON.stringify(errors),
-      });
+    } catch (err) {
+      if (err.endpoint === `retryGraphQL timeout`) {
```
mknoedel, line 626: I'm not a huge fan of using the endpoint in this way though there isn't a great way to find this timeout error otherwise.

mknoedel (approved):
Makes sense to me. Could consider keying off the Error message rather the endpoint when determining if it is a timeout error or not.

mknoedel (approved):

PR #137, opened 2022-01-11
Added concurrency to collabs and team-repos

kevincasey1222:
Here is some concurrency for the longer GitHub integration steps, `fetch-team-repos` and `fetch-collaborators`. It works well on my test account, but with so little data, I'm not sure if the speed gains I'm seeing are meaningful (10 seconds instead of 14 seconds). 

Can anyone run it against jupiterone-dev and see if the speed improvement is substantial? And also if there are any gotchas running it against that larger data set?


kevincasey1222:
Old dead PR

PR #140, opened 2022-01-25
Adjusting how teams are queried

kevincasey1222:
Hi @austinkelleher 

Update: this PR now solves all the problems


src/steps/teamMembers.ts
```diff
@@ -25,68 +27,81 @@ export async function fetchTeamMembers({
...
+  if (!allTeamNames) {
+    throw new IntegrationMissingKeyError(
+      `Expected teams.ts to have set ${GITHUB_ALL_TEAM_NAMES} in jobState.`,
+    );
+  }
+
   await jobState.iterateEntities(
     { _type: GITHUB_TEAM_ENTITY_TYPE },
     async (teamEntity: TeamEntity) => {
-      await apiClient.iterateTeamMembers(teamEntity, async (user) => {
```
kevincasey1222, line 31: Lines 31 onward were not really changed - we're just seeing GitHub respond to spacing changes.

src/steps/teamRepos.ts
```diff
@@ -22,43 +24,56 @@ export async function fetchTeamRepos({
...
+  if (!allTeamNames) {
+    throw new IntegrationMissingKeyError(
+      `Expected teams.ts to have set ${GITHUB_ALL_TEAM_NAMES} in jobState.`,
+    );
+  }
+
   await jobState.iterateEntities(
     { _type: GITHUB_TEAM_ENTITY_TYPE },
     async (teamEntity: TeamEntity) => {
-      await apiClient.iterateTeamRepos(teamEntity, async (teamRepo) => {
```
kevincasey1222, line 28: Lines in this file after this were not really changed - we're just seeing GitHub respond to spacing changes.

src/util/tuneTeamQueries.ts
```diff
@@ -0,0 +1,43 @@
...
+  slug: string,
+  teamNames: string[],
+  query: string,
+): string {
+  const ascSortPosition = findTeamSlugPositionInAscQuery(slug, teamNames);
+  const desSortPosition = calculateDesQueryPosition(
+    teamNames.length,
+    ascSortPosition,
+  );
+  if (ascSortPosition <= desSortPosition) {
```
austinkelleher, line 35: This seems like a very clever use of the GitHub API and took me a minute to really understand what was going on here. I'm worried that this is error prone. It seems to be making a lot of assumptions about the order of data returned by the GitHub API. 

Thinking further...Do we have a long-term approach that makes less assumptions?

austinkelleher (commented):
This seems like a

src/util/tuneTeamQueries.ts
kevincasey1222: My research yesterday led me to realize that this issue is foundational to the GraphQL spec. Fields are searched on a "contains" rather than "equals" basis. So we either query for extra teams (like we are doing today), which always runs a small but nonzero risk that we didn't ask for enough, or we get clever. 

Or we rewrite the pattern to pull all teams at once, which is how it used to be. But then we have the cursors problem.

The long-term perfect solution is perfect cursor handling, of course.

src/util/tuneTeamQueries.ts
```diff
@@ -0,0 +1,43 @@
...
+    // the comma in the substitution is important to not change the subquery
+    return query.replace('first: 1,', `first: ${ascSortPosition},`);
+  } else {
+    return query
+      .replace('first: 1,', `first: ${desSortPosition},`)
+      .replace('direction: ASC', 'direction: DESC');
```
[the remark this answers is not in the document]
kevincasey1222: Good point

kevincasey1222:
Okay. It was not simple to analyze, but the final fix was simple (though hacky). Thanks for your fresh eyes on that - the cursor expectations were blinding us to that much simpler GraphQL query, which actually solves the problem the right way. Just needed to hack the cursor system. 

So, this update is simple and solves the problem for all customers, with no regression. Also improves performance for everyone.

src/util/tuneTeamQueries.ts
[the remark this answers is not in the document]
kevincasey1222: But we ended up abandoning this approach.

austinkelleher (approved):

PR #144, opened 2022-01-28
Improve tests

kevincasey1222:
@austinkelleher @aiwilliams @ndowmon 

Hi guys! Here's a road I'm going down to improve tests in GitHub, based on work we've done in graph-aws.

The motive here is that a bad bug almost made it to production because it wasn't clear that our tests weren't thorough. And that was because the many steps of GitHub were being tested in one huge test, which is hard to analyze (and takes too long to run anyway).

The draft version of this PR just has a couple of the new tests (Accounts and Apps), and my intent is to convert the rest of the `index.test.ts` tests today or this weekend (and then delete `index.test.ts`. But I'm putting the PR up now in case anyone wants to comment or adjust my course before I do it for the other 12 steps.


src/steps/apps.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  let apps;
+  let context;
+
+  test('execution and snapshot', async () => {
+    recording = setupGithubRecording({
+      directory: __dirname,
+      name: 'apps', //redaction of headers is in setupGithubRecording
+    });
+    sanitizeConfig(integrationConfig);
+    integrationConfig.installationId = 17214088; //this is the id the recordings are under
```
ndowmon, line 28: FYI - in Azure, we use a function arg for polly `options: { matchRequestsBy: { url: { pathname: () => ...`, which allows us to replace whatever the config's `installationId` is with a constant string (sort of like when we replaced the body of `AssumeRole` requests with the string `"AssumeRole"` in the `graph-aws` project). 

It's a little less brittle than hard-coding the installationId, since it would theoretically allow recordings to match irrespective of the `installationId`. It can be sort of a complicated setup though so I'm not necessarily saying you should do this, just want to make you aware.

src/steps/account.test.ts
```diff
@@ -0,0 +1,75 @@
...
+      },
+    });
+  });
+
+  test('jobState includes DATA_ACCOUNT_ENTITY', async () => {
+    const entityFromConstant = await context.jobState.getData(
+      DATA_ACCOUNT_ENTITY,
+    );
+    expect(entityFromConstant).toEqual(accounts[0]);
+  });
```
ndowmon, line 74: Why not just do all of these in the same test? I'm not a huge fan of using global data that's instantiated inside of a single test.

src/steps/account.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    }).toMatchSnapshot();
+
+    accounts = context.jobState.collectedEntities.filter((e) =>
+      e._class.includes('Account'),
+    );
+  });
+
+  test('schema match', () => {
+    expect(accounts.length).toEqual(1);
+    expect(accounts).toMatchGraphObjectSchema({
```
ndowmon, line 50: Would you mind moving the schemas into `constants.ts`? See https://github.com/JupiterOne/graph-aws/pull/27#discussion_r794060916

src/steps/account.test.ts
```diff
@@ -0,0 +1,75 @@
...
+      numCollectedEntities: context.jobState.collectedEntities.length,
+      numCollectedRelationships: context.jobState.collectedRelationships.length,
+      collectedEntities: context.jobState.collectedEntities,
+      collectedRelationships: context.jobState.collectedRelationships,
+      encounteredTypes: context.jobState.encounteredTypes,
+    }).toMatchSnapshot();
+
+    accounts = context.jobState.collectedEntities.filter((e) =>
+      e._class.includes('Account'),
+    );
```
ndowmon, line 45: Whenever filtering `collectedEntities` or `collectedRelationships`, you should always use `_type`, since you can do a direct match:

```ts
context.jobState.collectedEntities.filter((e) => e._type === 'github_account');
```

That said, you should _only_ have `github_account` entities in this collection. 

```suggestion
    accounts = context.jobState.collectedEntities;
```

src/steps/apps.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    apps = context.jobState.collectedEntities.filter((e) =>
+      e._class.includes('Application'),
+    );
+    expect(apps.length).toBeGreaterThan(0);
+
+    const relationships = context.jobState.collectedRelationships.filter((e) =>
+      e._type.includes(GITHUB_ACCOUNT_APP_RELATIONSHIP_TYPE),
+    );
+    expect(relationships.length).toBeGreaterThan(0);
```
ndowmon, line 58: Similar to above:

```suggestion
    apps = context.jobState.collectedEntities;
    expect(apps.length).toBeGreaterThan(0);

    const relationships = context.jobState.collectedRelationships;
    expect(relationships.length).toBeGreaterThan(0);
```

src/steps/account.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        required: ['accountId'],
+      },
+    });
+  });
+
+  test('jobState includes DATA_ACCOUNT_ENTITY', async () => {
+    const entityFromConstant = await context.jobState.getData(
+      DATA_ACCOUNT_ENTITY,
+    );
+    expect(entityFromConstant).toEqual(accounts[0]);
```
ndowmon, line 73: Haven't ever thought about testing this, but I like the idea!

aiwilliams (commented):
Please take a look at https://github.com/JupiterOne/graph-aws/pull/27#issuecomment-1024684472.

src/steps/account.test.ts
kevincasey1222: Okay, cool. I've updated this to be a single test, and to use the new pattern Nick created with `executeStepWithDependencies`

src/steps/account.test.ts
kevincasey1222: Done!

kevincasey1222:
Dear potential reviewers... don't panic! Although there are 66 files changes, more than half of that is recordings and snapshots. Here's what changed:

1) `index.test.ts` has a massive yet still incomplete test. It's gone and replaced by very complete tests for each execution handler
2) I used a new pattern from @ndowmon , `executeStepWithDependencies`, to handle step dependency testing.
3) Many constants for entity type and class were moved into a single hierarchical object GithubEntities, which contains type, class, and schema for the entities we're testing. This was actually the only change that touched production code.

We needed to improve GitHub testing to prepare for the rewrite of nested resource cursors, which has great potential to break things. Thorough tests pave the way for that.

aiwilliams (approved):
This looks really, really good @kevincasey1222! Love the smaller, more focused tests. I added comments throughout, though I don't think any of them block this PR since as you pointed out, nothing is changing much in terms of production behavior. I think there is a real need for raw data patterns to improve across the board, which @austinkelleher is tracking.

PR #153, opened 2022-02-22
INT 2588 GitHub Enterprise Server Support

VDubber:
## This changeset adds support for GitHub Enterprise Server. 

- Updated .env to allow for customer base URL
- Added validation for the provided URL
- Rate limiting may not be enabled on GHE Server, adjusted to allow response to not contain those values.
- Updated development.md to include all permissions that are required.
- Updated clients to use the provided base URL.
- Updated converters to pull from configured base URL instead of hard coding to github.com

If you desired to run this against a GHE Server instance: https://ec2-34-226-194-193.compute-1.amazonaws.com/


.prettierignore
```diff
@@ -1,4 +1,5 @@
 dist
 coverage/
 .j1-integration
+.j1-integration-cache
```
ndowmon, line 4: Is this directory gitignored in `integration-template`? If not, can you open a PR there?

src/config.ts
```diff
@@ -72,6 +76,13 @@ export interface IntegrationConfig extends IntegrationInstanceConfig {
    * if this var is specified.
    */
   githubAppDefaultLogin: string;
+
+  /**
+   * Specifies the base URL to use when ingesting data.
+   * Allows for api.github.com or a self-hosted GitHub Enterprise
+   * Server to be used.
+   */
+  githubApiBaseUrl: string;
```
ndowmon, line 85: Do you think this should be optional? Assume `api.github.com` unless provided by the user?

src/validateInvocation.test.ts
```diff
@@ -2,40 +2,64 @@ import {
...
-    IntegrationValidationError,
-  );
 });
 
-it('auth error', async () => {
-  const recording = setupRecording({
-    directory: '__recordings__',
-    name: 'client-auth-error',
+describe('#validateInvocation', () => {
+  it('throws error if invalid url is supplied', async () => {
```
ndowmon, line 30: If you're up for it, would you be willing to update these `it()`s to `test()`s just to be consistent with the rest of the project?

docs/development.md
```diff
@@ -40,9 +40,22 @@ you'll need to do the following:
     - Webhook secret: `development`
     - At the bottom, under `Where can this GitHub App be installed?`, select
       `Any account`.
-    - Set repo permissions. Beside the Metadata Permissions always granted, the
-      App only needs Read Only for Repository Metadata and Organization Members
-      at this time.
+    - Set permissions as follows:
+
+      **Repository Permissions**
```
ndowmon, line 45: 🎉

ndowmon (dismissed):

.prettierignore
VDubber: Sure thing. I'll add it to .gitignore as well.

src/config.ts
VDubber: I tried initial to go that way but ran into an issue. I'm just realizing how to get this to work. 👍 
Loading....

src/config.ts
```diff
@@ -38,9 +38,6 @@ export const instanceConfigFields: IntegrationInstanceConfigFieldMap = {
   installationId: {
     type: 'string', //should be a number, but that's not an option in the SDK
   },
-  githubApiBaseUrl: {
-    type: 'string',
```
ndowmon: I feel like it'd be nice if the instance config field map allowed an `optional: true` property, so you can still add `githubApiBaseUrl` to this map... maybe something to think about adding in the future... 🤔

ndowmon (approved):
Looks awesome! Thanks for getting this set up for GHE!

src/config.ts
VDubber: Agreed! I expected it to have that ability.

VDubber: Create a Jira ticket to capture this thought

PR #159, opened 2022-03-09
Pagination Refactor - Draft

VDubber:
- Introduces new pagination system for pull requests. Handles nested pagination needs of inner resources.
- Calculates overall query cost.
- Respects rate limits
- Handles token refreshes

### DIFF
--- a/src/client/GraphQLClient/client.ts
+++ b/src/client/GraphQLClient/client.ts
@@ -34,6 +34,10 @@ import {
   ResourceMap,
   ResourceMetadata,
 } from './types';
+import PullRequestsQuery from './pullRequestQueries/PullRequestsQuery';
+import { createQueryExecutor } from './CreateQueryExecutor';
+
+const FIVE_MINUTES_IN_SECS = 300000;
 
 export class GitHubGraphQLClient {
   private readonly graphqlUrl: string;
@@ -122,7 +126,42 @@ export class GitHubGraphQLClient {
     }
   }
 
+  public async iteratePullRequestsV2(
+    repository: { fullName: string; public: boolean },
+    lastExecutionTime: string,
+    iteratee: ResourceIteratee<PullRequest>,
+  ): Promise<QueryResponse> {
+    const executor = createQueryExecutor(this, this.logger);
+
+    const { rateLimitConsumed } = await PullRequestsQuery.iteratePullRequests(
+      repository,
+      lastExecutionTime,
+      iteratee,
+      executor,
+    );
+
+    return {
+      rateLimitConsumed,
+    };
+  }
+
   /**
+   * Performs GraphQl request.
+   * Handles:
+   *    - token management (refreshes 5 minutes before expiring)
+   *    - rate limit management
+   * @param queryString
+   * @param queryVariables
+   */
+  public async query(queryString: string, queryVariables) {
+    if (this.tokenExpires - FIVE_MINUTES_IN_SECS < Date.now()) {
+      await this.refreshToken();
+    }
+    return await this.retryGraphQLV2(queryString, queryVariables);
+  }
+
+  /**
+   * @Deprecated use iteratePullRequestsV2
    * Iterates through a search request for Pull Requests while handling
    * pagination of both the pull requests and their inner resources.
    *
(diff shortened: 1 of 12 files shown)


src/client/GraphQLClient/client.ts
```diff
@@ -609,6 +648,110 @@ export class GitHubGraphQLClient {
...
+   * @private
+   */
+  private async retryGraphQLV2(queryString: string, queryVariables) {
+    const { logger } = this;
+
+    //queryWithRateLimitCatch will be passed to the retry function below
+    const queryWithRateLimitCatch = async () => {
+      let response;
+      try {
+        response = await this.graph(queryString)(queryVariables);
```
VDubber, line 664: Allows the token to be refreshed.

src/client/GraphQLClient/pullRequestQueries/PullRequestsQuery.ts
```diff
@@ -0,0 +1,266 @@
...
+              ${labelsQuery} 
+            }
+          }
+          pageInfo {
+            endCursor
+            hasNextPage
+          }
+        }
+        ...rateLimit
+      }`;
```
zemberdotnet, line 74: I'm pretty new with `graph-ql`. Is this the query language for `graph-ql`? Just curious looks very similar to `ts`.

src/client/GraphQLClient/client.ts
```diff
@@ -609,6 +650,110 @@ export class GitHubGraphQLClient {
     return aggregatedResources;
   }
 
+  /**
+   *
+   * @param queryString
+   * @param queryVariables
+   * @private
+   */
+  private async retryGraphQLV2(queryString: string, queryVariables) {
```
VDubber, line 659: Additional refactoring to come on this method!

src/client/GraphQLClient/pullRequestQueries/PullRequestsQuery.ts
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    if (
+      utils.innerResourcePaginationRequired(pullRequest) &&
+      utils.hasRepoOwnerAndName(pullRequest)
+    ) {
+      const repoOwnerAndName = utils.findRepoOwnerAndName(pullRequest);
+      // Add pull request to queue allowing process to continue without
+      // pausing for subsequent requests. This path is not as common.
+      onInnerResourceQueryRequired({
```
zemberdotnet, line 188: I'm a little confused on this function. Does this build and execute another query to get the inner resource?

zemberdotnet: Never mind, I got to the next function and it made sense.

src/client/GraphQLClient/client.ts
```diff
@@ -609,6 +650,110 @@ export class GitHubGraphQLClient {
...
+          return;
+        }
+
+        // don't keep trying if it's not going to get better
+        if (err.retryable === false || err.status === 403) {
+          logger.warn(
+            { attemptContext, err },
+            `Hit an unrecoverable error when attempting to query GraphQL. Aborting.`,
+          );
+          attemptContext.abort();
```
adam-in-ict, line 731: So a 403 is the only error we abort and give up?  Would we also want to add a 500 error catch at a minimum just to catch the edge case where the server is falling on its face?  I don't imagine it's a huge issue if we retry in that instance, but it'd let us fail faster.

src/client/GraphQLClient/pullRequestQueries/SinglePullRequestQuery.ts
```diff
@@ -0,0 +1,230 @@
...
+const MAX_REQUESTS_NUM = 100;
+
+/**
+ * Builds the leanest query possible
+ * based on the provided queryState.
+ *
+ * @param queryParams
+ * @param queryState
+ */
+export const buildQuery: BuildQuery<QueryParams, QueryState> = (
```
zemberdotnet, line 34: Why does the single pull request need a separate query? It seems like it is used to get more detailed information for each PR? Is that the case?

src/client/GraphQLClient/types/query.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export type BuildQuery<P, S extends BaseQueryState> = (
+  queryParams: P,
+  queryState?: S,
+) => ExecutableQuery;
+
+export type ProcessedData<S extends BaseQueryState> = {
+  resource;
+  queryState: S;
+};
```
zemberdotnet, line 26: Really appreciate the types functions. I think this will help a lot in extending this pattern in the future.

src/client/GraphQLClient/client.ts
```diff
@@ -34,6 +34,10 @@ import {
   ResourceMap,
   ResourceMetadata,
 } from './types';
+import PullRequestsQuery from './pullRequestQueries/PullRequestsQuery';
+import { createQueryExecutor } from './CreateQueryExecutor';
+
+const FIVE_MINUTES_IN_SECS = 300000;
```
adam-in-ict, line 40: Do you mean milliseconds?

VDubber: Yes!

src/client/GraphQLClient/pullRequestQueries/SinglePullRequestQuery.ts
VDubber: Nested pagination is messy/complicated. So if a PR has inner resources that need pagination, this query is used instead where the pagination isn't nested so deep.

VDubber: It'd be good for me to explain that in a doc.

src/client/GraphQLClient/client.ts
adam-in-ict: 👍  In that case, I'd probably just call it a nitpick item to recommend changing the name to call out milli.  Not a big deal.  I just didn't want you to potentially be refreshing 1000x earlier than you'd intended.

src/client/GraphQLClient/pullRequestQueries/PullRequestsQuery.ts
VDubber: Yep! There technically is `ts` in the middle of the query, using string interpolation. But it does have a "comfortable" feel to it. Here's where I had to start if you are interested: https://graphql.org/learn/queries/

test/recording.ts
```diff
@@ -18,6 +24,7 @@ export function setupGithubRecording(
       redact(entry);
     },
     options: {
+      mode: recordingEnabled ? 'record' : 'replay',
```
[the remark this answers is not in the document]
VDubber: Yeah! Forces it to record instead of replay.

src/client/GraphQLClient/client.ts
VDubber: Ha! Thanks. I'll get it cleaned up.

src/client/GraphQLClient/client.ts
VDubber: I copy/pasted this method a changed a very small portion of it. It seems to be working as I would hope so I'll go back and clean this up. I'll also consider the scenario you presented here.

PR #166, opened 2022-03-29
INT-2824 Improved error handling

VDubber:
- Identified valid errors
- Introduced clear error handling
- Added tests
- Upgrade graphql.js
- Added example errors to README.md


src/util/validateGraphQLReponse.ts
```diff
@@ -1,70 +0,0 @@
-import {
-  IntegrationLogger,
-  IntegrationProviderAPIError,
-} from '@jupiterone/integration-sdk-core';
-
-export default function validateGraphQLResponse(
```
VDubber, line 6: Rate limit errors come through a rejected promise instead of the resolved promise.

kevincasey1222 (approved):
I'm not a code owner so my review doesn't let you merge... but having worked on this integration a bunch last year, I just wanted to say that your work looks nicely thought through and I appreciate that you took the time to unpack all the layers of errors that can present in this integration. From my memory of how it works, I don't see any concerns with your work here - it looks smart.

adam-in-ict (approved):
My approval seems somewhat superfluous if Kevin has already looked over it, but I also think it looks good to go.  Appreciate all the work documenting the different JSON structures for different possible errors. 💯

PR #168, opened 2022-03-31
Add partial flag to issues entity & relationship

VDubber:
When we query for GitHub Issues, we base the ingestion on the last successful integration run. This allows us to only query for new data that needs to be added into the graph. The issues entity and relationships are not currently marked as partial, which triggers the differ to remove all existing Issues.


kevincasey1222 (approved):
Looks great! 

You might also consider updating the notes at https://github.com/JupiterOne/graph-github/blob/main/docs/jupiterone.md to reflect this change in behavior. In the "How It Works" section there's a note mentioning that PRs accumulate... we could point out that Issues now do also.

kevincasey1222 (approved):
Perfect! Now you just need someone who is not a contractor to approve. :)

adam-in-ict (approved):

PR #171, opened 2022-04-18
Fix OrganizationQuery.ts to handle when rateLimit is not included

VDubber:
- Fix OrganizationQuery.ts to handle when rateLimit is not included in response
   - Possible for GitHub Enterprise server configuration
- Added test without rateLimit in response
- Removed rateLimit from IssuesQuery.test.ts to verify functionality.

### DIFF
--- a/src/client/GraphQLClient/organizationQueries/OrganizationQuery.ts
+++ b/src/client/GraphQLClient/organizationQueries/OrganizationQuery.ts
@@ -45,7 +45,7 @@ const processResponseData = (responseData): OrganizationResults => {
   return {
     rateLimit: {
       ...rateLimit,
-      totalCost: rateLimit.cost,
+      totalCost: rateLimit?.cost ?? 0,
     },
     organization,
   };
--- a/src/client/OrganizationAccountClient.ts
+++ b/src/client/OrganizationAccountClient.ts
@@ -79,6 +79,11 @@ export default class OrganizationAccountClient {
 
     this.authorizedForPullRequests = true;
     this.v3RateLimitConsumed = 0;
+
+    this.logger.info(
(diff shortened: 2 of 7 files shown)


src/client/GraphQLClient/organizationQueries/OrganizationQuery.ts
```diff
@@ -45,7 +45,7 @@ const processResponseData = (responseData): OrganizationResults => {
   return {
     rateLimit: {
       ...rateLimit,
-      totalCost: rateLimit.cost,
+      totalCost: rateLimit?.cost ?? 0,
```
VDubber, line 48: This was _the_ fix

src/client/OrganizationAccountClient.ts
```diff
@@ -79,6 +79,11 @@ export default class OrganizationAccountClient {
 
     this.authorizedForPullRequests = true;
     this.v3RateLimitConsumed = 0;
+
+    this.logger.info(
+      { login: options.login, baseUrl: options.baseUrl },
```
austinkelleher, line 84: What is the value of `login` look like? According to the comment above:

```ts
    /**
     * The login of the GitHub account, a required parameter of some endpoints.
     */
```

It's still not immediately clear to me from the comment though 😄

austinkelleher (approved):

src/client/OrganizationAccountClient.ts
VDubber: Yeah, GitHub had to make it weird. login === org id/name. It's used in a lot of queries.

I am logging these values to double check that a customer's login value is correctly determined.
`login: installation?.account?.login ?? this.config.githubAppDefaultLogin,` from src/client.ts:469

PR #175, opened 2022-04-22
INT-3483 - fix sentry error

VDubber:
`iteratePullRequests: Cannot read property 'hasNextPage' of undefined`

- Add null dot operator to hasNextPage
- Removed old error handling that was hiding details.
https://sentry.io/organizations/jupiterone/issues/2548740334/?project=5880207&query=is%3Aunresolved


austinkelleher (dismissed):
Did you validate whether this could have occurred due to a step dependency failing? Do you know under what circumstances these values could actually be undefined?

src/client/GraphQLClient/pullRequestQueries/SinglePullRequestQuery.test.ts
```diff
@@ -38,6 +41,33 @@ describe('SinglePullRequestQuery', () => {
...
+      const queryState = {
+        commits: {
+          endCursor: 'MQ',
+          hasNextPage: false,
+        },
+        labels: {
+          endCursor: 'Y3Vyc2==',
+          hasNextPage: true,
+        },
+        reviews: undefined,
```
VDubber, line 57: @austinkelleher This is the scenario. The first query will always include pageInfo for commits, labels, and reviews. But as soon as the inner resource has no more, that portion of the query is removed to keep things as cheap (rate limiting) as possible.

src/client/GraphQLClient/pullRequestQueries/SinglePullRequestQuery.ts
```diff
@@ -41,27 +42,44 @@ export const buildQuery: BuildQuery<QueryParams, QueryState> = (
         $repoOwner: String!
         $maxLimit: Int!
         ${
-          queryState?.commits.hasNextPage !== false
+          queryState?.isInitialQuery ||
```
VDubber, line 45: This boolean logic was confusing. I've clarified it to allow hasNextPage to only be about pagination and not also about the initial query state.

austinkelleher (approved):

PR #179, opened 2022-04-27
INT 3534 move to octokit graphql

VDubber:
To prevent low level errors, we moved away from graphql.js package to octokit/graphql.js package. This appears to be a more robust and well maintained package from GitHub. 
We lost `fragments` feature. String interpolation is essentially what it was doing and those represent a bulk of the changes.
Snapshots were updated as well since the fragments feature didn't execute prior to snapshotting the query.

The new package handles the baseUrl differently. It appends `/graphql` for us.

### DIFF
--- a/src/client/GraphQLClient/client.ts
+++ b/src/client/GraphQLClient/client.ts
@@ -1,4 +1,4 @@
-import { graphql as ocktokitGraphQl } from '@octokit/graphql';
+import { graphql as octokitGraphQl } from '@octokit/graphql';
 
 import {
   IntegrationLogger,
@@ -90,11 +90,11 @@ export class GitHubGraphQLClient {
     authClient: Octokit,
   ) {
     this.graphqlUrl = graphqlUrl;
-    this.graph = ocktokitGraphQl.defaults({
+    this.graph = octokitGraphQl.defaults({
       baseUrl: this.graphqlUrl,
       headers: {
         'User-Agent': 'jupiterone-graph-github',
-        authorization: `token ${token}`,
+        Authorization: `token ${token}`,
       },
     });
     this.tokenExpires = tokenExpires;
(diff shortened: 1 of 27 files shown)


src/client/GraphQLClient/client.ts
```diff
@@ -90,14 +90,13 @@ export class GitHubGraphQLClient {
     authClient: Octokit,
   ) {
     this.graphqlUrl = graphqlUrl;
-    this.graph = graphql(this.graphqlUrl, {
+    this.graph = ocktokitGraphQl.defaults({
+      baseUrl: this.graphqlUrl,
       headers: {
         'User-Agent': 'jupiterone-graph-github',
-        Authorization: `token ${token}`,
+        authorization: `token ${token}`,
```
adam-in-ict, line 97: I'm assuming this is still case insensitive like all HTTP headers.  Is that a correct assumption?  Just wanted to make sure it was a preference choice and not an indicator of a special gotcha when switching to ocktokit.

src/client/GraphQLClient/client.ts
```diff
@@ -90,14 +90,13 @@ export class GitHubGraphQLClient {
     authClient: Octokit,
   ) {
     this.graphqlUrl = graphqlUrl;
-    this.graph = graphql(this.graphqlUrl, {
+    this.graph = ocktokitGraphQl.defaults({
```
adam-in-ict, line 93: nit:  octokitGraphQl instead of ocktokitGraphQl

src/client/GraphQLClient/client.ts
```diff
@@ -1,4 +1,4 @@
-import graphql, { GraphQLClient } from 'graphql.js';
+import { graphql as ocktokitGraphQl } from '@octokit/graphql';
```
adam-in-ict, line 1: nit:  octokitGraphQl instead of ocktokitGraphQl

adam-in-ict (dismissed):

src/client/GraphQLClient/client.ts
VDubber: I followed their documentation which uses `authorization`... But I just tested, either works just fine 👍

adam-in-ict: Sounds good! Just curious. Thanks!

adam-in-ict (approved):

PR #181, opened 2022-04-28
INT 3556 octokit errors

VDubber:
The previous graphql client unwrapped the error, only passing error.errors.
The new client returns the entire error. So logic need to be rewired. 
With the entire error, we are also able to return partial data. 

Since errors are hard to manually test, please review the docs [here](https://github.com/octokit/graphql.js/#errors) to see if I missed anything.

### DIFF
--- a/src/client/GraphQLClient/client.ts
+++ b/src/client/GraphQLClient/client.ts
@@ -391,10 +391,10 @@ export class GitHubGraphQLClient {
               // Partial data can be included in errors.
               ...error.data,
             };
-          } else {
-            throw error;
           }
         }
+
+        throw error;
       }
     };
 
--- a/src/client/GraphQLClient/client.ts
+++ b/src/client/GraphQLClient/client.ts
@@ -1,4 +1,7 @@
-import { graphql as octokitGraphQl } from '@octokit/graphql';
+import {
+  graphql as octokitGraphQl,
+  GraphqlResponseError,
+} from '@octokit/graphql';
 
 import {
   IntegrationLogger,
@@ -369,33 +372,27 @@ export class GitHubGraphQLClient {
           'Attempting GraphQL request',
         );
         return await this.graph(queryString, queryVariables);
-      } catch (err) {
+      } catch (error) {
         logger.debug(
-          { queryString, queryVariables, timeoutRetryAttempt, err },
+          { queryString, queryVariables, timeoutRetryAttempt, error },
           'GraphQL request failed.',
         );
 
-        // Add queryString & queryVariables to error.
-        if (Array.isArray(err) && err.length > 0) {
-          err[0].queryString = queryString;
-          err[0].queryVariables = queryVariables;
-        } else {
-          err.queryString = queryString;
-          err.queryVariables = queryVariables;
-        }
-
-        // Handle pre-retry logic
-        // If resource can't be found, or is not accessible,
-        // continue with processing.
-        if (
-          handleNotFoundErrors(err, logger) ||
-          handleForbiddenErrors(err, logger)
-        ) {
-          return {
-            rateLimit: {},
-          };
-        } else {
-          throw err;
+        if (error instanceof GraphqlResponseError) {
+          // Handle pre-retry logic
+          // If resource can't be found, or is not accessible,
+          // continue with processing.
+          if (
+            handleNotFoundErrors(error.errors, logger) ||
+            handleForbiddenErrors(error.errors, logger)
+          ) {
+            return {
+              rateLimit: {},
+              ...error.data,
+            };
+          } else {
+            throw error;
+          }
         }
       }
     };
--- a/src/client/GraphQLClient/client.ts
+++ b/src/client/GraphQLClient/client.ts
@@ -388,6 +388,7 @@ export class GitHubGraphQLClient {
           ) {
             return {
               rateLimit: {},
+              // Partial data can be included in errors.
               ...error.data,
             };
           } else {
--- a/src/client/GraphQLClient/errorHandlers.ts
+++ b/src/client/GraphQLClient/errorHandlers.ts
@@ -75,22 +75,22 @@ export const retryErrorHandle = async (
       error.errors?.some((e) => e.type === 'RATE_LIMITED')
     ) {
       logger.info({ attemptContext, error }, 'Rate limiting message received.');
-    } else if (error.message?.includes('Bad credentials')) {
-      logger.info({ error }, 'Bad credentials: Refreshing token.');
-      await refreshToken();
-    } else if (error.message?.includes('exceeded a secondary rate limit')) {
-      logger.info(
-        { attemptContext, error },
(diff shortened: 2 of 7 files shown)


src/client/GraphQLClient/client.ts
```diff
@@ -369,33 +372,28 @@ export class GitHubGraphQLClient {
...
+          { queryString, queryVariables, timeoutRetryAttempt, error },
           'GraphQL request failed.',
         );
 
-        // Add queryString & queryVariables to error.
-        if (Array.isArray(err) && err.length > 0) {
-          err[0].queryString = queryString;
-          err[0].queryVariables = queryVariables;
-        } else {
-          err.queryString = queryString;
```
VDubber, line 383: No longer valuable since the error includes these values.

src/client/GraphQLClient/client.ts
```diff
@@ -369,33 +372,28 @@ export class GitHubGraphQLClient {
...
-        if (
-          handleNotFoundErrors(err, logger) ||
-          handleForbiddenErrors(err, logger)
-        ) {
-          return {
-            rateLimit: {},
-          };
-        } else {
-          throw err;
+        if (error instanceof GraphqlResponseError) {
```
VDubber, line 381: This is from the [docs](https://github.com/octokit/graphql.js/#errors)

src/client/GraphQLClient/client.ts
zemberdotnet: Nice 👍

src/client/GraphQLClient/utils.ts
```diff
@@ -18,9 +18,11 @@ const responseToResource = (node) => {
...
   return {
     ...node,
-    commits: node.commits?.nodes?.map((node) => node.commit) ?? [],
-    reviews: node.reviews?.nodes ?? [],
-    labels: node.labels?.nodes ?? [],
+    commits:
+      node.commits?.nodes?.filter((node) => node).map((node) => node.commit) ??
+      [],
+    reviews: node.reviews?.nodes?.filter((node) => node) ?? [],
+    labels: node.labels?.nodes?.filter((node) => node) ?? [],
```
zemberdotnet, line 25: Does the client get the response for this in a different format? What caused this change?

src/client/GraphQLClient/errorHandlers.ts
```diff
@@ -1,26 +1,21 @@
...
-  errors = Array.isArray(errors) ? errors : [errors];
-
-  if (errors?.every((error) => error.type === type)) {
+const handleTypeErrors = (
+  errors: GraphQlQueryResponse<never>['errors'],
+  logger,
+  type: string,
+): boolean => {
+  if (Array.isArray(errors) && errors?.every((error) => error.type === type)) {
+    logger.info(
```
zemberdotnet, line 12: Could you add a comment here explaining what this is for? If every error is `type` we ignore them?

src/client/GraphQLClient/errorHandlers.ts
```diff
@@ -1,26 +1,21 @@
...
+      { errors, type },
+      'The error was found and ignored because of the type.',
+    );
     errors.forEach((error) => {
       if (error.message) {
-        logger.warn(error.message);
+        logger.debug(error.message);
       }
```
zemberdotnet, line 19: Did we downgrade this to `debug` because it's not as impactful as we thought?

src/client/GraphQLClient/errorHandlers.ts
```diff
@@ -62,24 +57,25 @@ export const handleForbiddenErrors = (
...
  */
 export const retryErrorHandle = async (
-  error: GraphQLError[] | GraphQLError,
+  error,
   logger: IntegrationLogger,
   attemptContext: AttemptContext,
   refreshToken: () => Promise<void>,
 ) => {
-  if (Array.isArray(error)) {
+  if (error instanceof GraphqlResponseError) {
```
zemberdotnet, line 65: Do we do nothing if the error isn't `GraphqlResponseError`?

src/client/GraphQLClient/client.ts
```diff
@@ -369,33 +372,28 @@ export class GitHubGraphQLClient {
...
+          // Handle pre-retry logic
+          // If resource can't be found, or is not accessible,
+          // continue with processing.
+          if (
+            handleNotFoundErrors(error.errors, logger) ||
+            handleForbiddenErrors(error.errors, logger)
+          ) {
+            return {
+              rateLimit: {},
+              // Partial data can be included in errors.
```
zemberdotnet, line 391: It looks like this was already here, so it's all good, but what type does `queryWithPreRetryErrorHandling` return? I see above that it returns
```ts
         return await this.graph(queryString, queryVariables);
```

Is this the same type as what is returned there?

src/client/GraphQLClient/utils.ts
VDubber: I saw an actual response where nodes[] was filled with all nulls. eg [null, null, null]. So I wanted to filter out undefined values.

src/client/GraphQLClient/errorHandlers.ts
VDubber: Yeah, this method is generic, below there are two cases in which we ignore the errors. 1. if the type = FORBIDDEN 2. if the type = NOT_FOUND.

They are safe to ignore and allow the ingestion to continue.

src/client/GraphQLClient/errorHandlers.ts
VDubber: Correct. It ended up just being the same message repeated 100s of times

src/client/GraphQLClient/errorHandlers.ts
VDubber: Thanks for catching this! Huge bug avoided!

src/client/GraphQLClient/client.ts
VDubber: Correct. rateLimit is returned in every response. Instead of using the headers, GH graphql allows for the rate limit info to be included in the response body.

src/client/GraphQLClient/errorHandlers.ts
VDubber: I'll add a comment

adam-in-ict: I can see why we'd ignore a NOT_FOUND error, but is there any concern for inadvertently masking missing ingesting data because our credentials aren't given enough permissions and repeatedly getting FORBIDDEN?

I'm not saying I think this code is incorrect as written.  I just wonder if we might want to put in a TODO to address the FORBIDDEN errors once we determine how we're going to report skipped items to users.

VDubber: The reporting feature would be great here.
In the errors I have seen, this comes up for only a portion of the response data, a nest part of the graph query. I'll get a todo in there

adam-in-ict (dismissed):

adam-in-ict (approved):

PR #183, opened 2022-04-29
Int 3568 Incorrect baseUrl

VDubber:


src/client/GraphQLClient/client.ts
```diff
@@ -49,30 +49,7 @@ export class GitHubGraphQLClient {
   private readonly graphqlUrl: string;
 
   /**
-   * A function generated by `graphql()` that is bound to authentication
```
VDubber, line 52: Stale. Referred to old graphql client

zemberdotnet (approved):

PR #189, opened 2022-05-10
INT-3401 action handler

VDubber:
To better understand the what & why, refer to this [confluence doc](https://jupiterone.atlassian.net/wiki/spaces/INT/pages/49578022/Project+Integration+Actions).

This PR adds the ability for the j1 nodejs client to call `invokeIntegrationAction(integrationInstanceId, action)` and have the code found in this PR build the requested entities and return them. 

Follow on PRs will add the ability for these entities to be ingested into the graph.


src/actions/partialIngestActionHandler.ts
```diff
@@ -0,0 +1,118 @@
...
+
+      const pullRequest = decomposePullRequestKey(entity._key);
+
+      return buildEntity(client, {
+        repoOwner: pullRequest.login,
+        repoName: pullRequest.repoName,
+        pullRequestNumber: pullRequest.pullRequestNumber,
+      });
+
+      // TODO: ingest entity and relationships
```
VDubber, line 46: Add date/initials/ticket

src/actions/actionExecutionHandler.ts
```diff
@@ -0,0 +1,46 @@
...
+  const { logger, instance, event } = context;
+  const { action } = event;
+
+  logger.debug({ integrationAction: action }, 'Handling integration action');
+
+  const client = getOrCreateApiClient(instance.config, logger);
+
+  await client.verifyAuthentication();
+
+  switch (action.name) {
```
adam-in-ict, line 34: Any concern not having a mechanism in this switch statement to handle (or at least log) action.name values we weren't expecting?  Or is that something that's typical and we really only want to pay attention to 'PARTIAL_INGEST'?

src/actions/partialIngestActionHandler.ts
```diff
@@ -0,0 +1,118 @@
...
+    } else {
+      invalidEntities.push(entity);
+    }
+  });
+
+  const entities = await Promise.all(pendingPromises);
```
adam-in-ict, line 52: We'd discussed on a totally separate and unrelated PR the fact that Promise.all will reject on any rejected promises in the array.  Are we ok with that?  If we upgraded the target lib to es2020 we'd be able to use Promise.allSettled() if that was a concern (it returns resolved as long as all promises are either fulfilled or rejected).

adam-in-ict (approved):

src/actions/partialIngestActionHandler.ts
VDubber: It'd be nice to return successfully generated entities and errors together. I'll see what I can do.

src/actions/actionExecutionHandler.ts
VDubber: I'll add a default case and throw an unsupported error. At the level, we only support PARTIAL_INGEST

src/actions/actionExecutionHandler.ts
```diff
@@ -40,7 +43,14 @@ export default async function actionExecutionHandler(
       return await partialIngestActionHandler(
         client,
         action.parameters.entities,
+        logger,
       );
     }
+    default: {
+      throw new IntegrationError({
+        code: 'UNKNOWN_ACTION_NAME',
+        message: 'The provided action.name is not unknown.',
```
adam-in-ict, line 52: I think you meant `The provided action.name is unknown.`. Otherwise looks 💯

adam-in-ict (approved):

src/actions/actionExecutionHandler.ts
VDubber: Dang. I'll catch it on the next round

PR #193, opened 2022-05-24
Int 383 fix&relationship

VDubber:
Fixes a undefined error and replaces direct relationship with mapped relationship

### DIFF
--- a/src/sync/converters.ts
+++ b/src/sync/converters.ts
@@ -1,34 +1,38 @@
 import {
-  setRawData,
-  parseTimePropertyValue,
-  RelationshipClass,
   createIntegrationEntity,
-  MappedRelationship,
-  RelationshipDirection,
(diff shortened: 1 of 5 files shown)


src/sync/converters.ts
```diff
@@ -527,7 +531,7 @@ type VulnerabilityAlertCweResponse = {
   description: string;
 };
 
-export function toCweEntity(cwe: VulnerabilityAlertCweResponse) {
+export function createCweEntity(cwe: VulnerabilityAlertCweResponse) {
   const cweNumber = cwe.cweId.replace(/^\D+/g, '');
   const cweId = cwe.cweId.toLowerCase();
```
austinkelleher, line 536: This `toLowerCase` is important to ensure that `_key` matches other integrations. 👍 👍

PR #197, opened 2022-06-13
Add Dependabot Alert config with state and severity filters.

VDubber:
- Add vuln state and severity filters.
- Vuln ingestion will be automatically be turned off by default. Users can enable at will
- Updated Vuln Query to use filters. State filter modifies GraphQL query, severity filter is performed in the client
- Log missing scopes to job event log


src/config.ts
```diff
@@ -154,6 +169,15 @@ export function sanitizeConfig(config: IntegrationConfig) {
       'https://api.github.com',
   );
 
+  config.enableDependabotAlerts =
+    config.enableDependabotAlerts ||
+    Boolean(process.env['ENABLE_DEPENDABOT_ALERTS']);
+
+  if (config.enableDependabotAlerts) {
```
adam-in-ict, line 176: nit:  Would it be helpful to add some or all of these new config values to the .env.example file?

adam-in-ict (dismissed):

src/config.ts
VDubber: Yes!

PR #199, opened 2022-06-15
Add support for Vuln Alert Ingestion for GHE Server <3.5               

VDubber:
When testing changes on 3.3.3 GHE Server, many errors were found. These changes include fixes for those errors plus an additional request to fetch server version data.


src/sync/converters.ts
```diff
@@ -461,20 +468,25 @@ export function createVulnerabilityAlertEntity(
...
+          ? data.state === RepositoryVulnerabilityAlertState.Open
+          : !data.dismissedAt,
         references: data.securityAdvisory?.references?.map((ref) => ref.url),
         public: data.securityAdvisory?.identifiers?.some(
           (identifier) => identifier.type === 'CVE',
         ),
         weblink: apiUrlToWebLink(
           baseUrl,
-          `/${data.repository.nameWithOwner}/security/dependabot/${data.number}`,
+          `/${data.repository.nameWithOwner}/security/dependabot/${
```
VDubber, line 480: I tried to determine what this URL would be for version < 3.5 but to enable dependabot requires us to connect our GH Server to a Cloud Enterprise account.

VDubber: Without the number, it will take the user to a list of their alerts.

src/steps/vulnerabilityAlertsSteps.test.ts
```diff
@@ -19,86 +19,88 @@ afterEach(async () => {
   await recording.stop();
 });
 
-test('fetchVulnerabilityAlerts exec handler', async () => {
```
VDubber, line 22: Wrapped the tests in a `describe`.

src/client/GraphQLClient/vulnerabilityAlertQueries/RepoVulnAlertsQuery.ts
```diff
@@ -125,6 +128,48 @@ const buildQuery: BuildQuery<QueryParams, QueryState> = (
...
+      EnterpriseFeatures.REPO_VULN_ALERT_FIELDS,
+      gheServerVersion,
+    )
+  ) {
+    fragments.additionalFields = 'fixReason\n fixedAt\n number\n state';
+  }
+
+  return fragments;
+};
+
```
adam-in-ict, line 172: I'm still getting my head wrapped around the use of fragments, but this looks like a great use case! 👍 💯

src/client.ts
```diff
@@ -82,10 +83,20 @@ export class APIClient {
...
+  /**
+   * Queries meta endpoint and saves the GHE Server version, if applicable.
+   */
+  private async fetchAndSetupMeta(): Promise<void> {
+    if (!this.accountClient) {
+      await this.setupAccountClient();
+    }
+
+    const meta = await this.accountClient.fetchMeta();
+    this.gheServerVersion = meta.installed_version ?? null;
```
adam-in-ict, line 98: I know we're logging the info on the overall `meta` object, but I almost wonder if it wouldn't be nice to have the GitHub Enterprise server version stored on a node in the ingested data somewhere too. Like maybe the account or environment node?  Not sure.  Might go against our overall data model.

adam-in-ict (approved):

src/client.ts
VDubber: Interesting thought. 🤔  Looked in `data-model` for a `version` property on a schema and didn't find much.

VDubber: Maybe I'll just add it to the account entity

PR #201, opened 2022-06-17
INT-4057 - Converting to mapped relationship

adam-in-ict:
Converting relationship between vuln and cwe/cve to be a mapped relationship.

### DIFF
--- a/src/steps/vulnerabilityAlertsSteps.ts
+++ b/src/steps/vulnerabilityAlertsSteps.ts
@@ -1,6 +1,5 @@
 import {
   createDirectRelationship,
-  Entity,
   IntegrationStep,
   IntegrationStepExecutionContext,
   RelationshipClass,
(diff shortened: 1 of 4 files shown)


src/steps/vulnerabilityAlertsSteps.ts
```diff
@@ -66,12 +59,7 @@ export async function fetchVulnAlerts({
         const cwes = vulnAlert.securityAdvisory?.cwes ?? [];
         await Promise.all(
           cwes.map(async (cwe) => {
-            const cweKey = cwe.cweId.toLowerCase();
-
-            let cweEntity = await jobState.findEntity(cweKey);
-            if (!cweEntity) {
-              cweEntity = await jobState.addEntity(createCweEntity(cwe));
-            }
```
zemberdotnet, line 74: Why do we not have to check to see if it exists anymore? Because it's now a mapped relationship?

zemberdotnet: I don't have my head completely around this, but is it not possible that duplicate relationships are created? Would that not trigger a duplicate key error?

adam-in-ict: I think previously the concern was multiple findings linked to one CVE or CWE.  So we had to make sure the entity itself was created only for the first relationship to it and all the others re-used it from the jobState.  We'll still be creating the same number of relationships as we were before, we're just relying on the mapper to create the target entity instead of adding it to this integration's jobState.

zemberdotnet: Ah, okay. That makes sense 😄

zemberdotnet (dismissed):

src/sync/converters.ts
```diff
@@ -561,24 +561,22 @@ export function createFindingCveRelationship(
...
-        _type: cveEntity._type,
-        _class: cveEntity._class,
-        id: cveEntity.id,
-        displayName: cveEntity.displayName,
-        name: cveEntity.name,
-        cvssScore: cveEntity.cvssScore,
-        webLink: cveEntity.webLink,
-        references: cveEntity.references,
-      },
+    target: {
```
VDubber, line 567: Now we know!  🎉

VDubber (approved):

zemberdotnet (approved):

PR #208, opened 2022-06-24
Checking for and handling duplicate PR keys

adam-in-ict:
Based on preliminary testing, if we receive a new PR into a repo while paginating, the results will shift and cause us to see a PR twice.

### DIFF
--- a/src/steps/pullRequests.ts
+++ b/src/steps/pullRequests.ts
@@ -94,9 +94,14 @@ export async function fetchPrs(
               teamMembersByLoginMap,
               usersByLoginMap!,
             );
-            const prEntity = (await jobState.addEntity(
-              pr,
-            )) as PullRequestEntity;
+            // If we receive a new PR into a repo while paginating, the
+            // results will shift and cause us to see a PR twice.
+            let prEntity = await jobState.findEntity(pr._key);
+            if (prEntity) {
+              logger.info({ key: pr._key }, 'PR already exists in job state.');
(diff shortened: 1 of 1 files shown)


src/steps/pullRequests.ts
```diff
@@ -94,9 +94,14 @@ export async function fetchPrs(
...
-              pr,
-            )) as PullRequestEntity;
+            // If we receive a new PR into a repo while paginating, the
+            // results will shift and cause us to see a PR twice.
+            let prEntity = await jobState.findEntity(pr._key);
+            if (prEntity) {
+              logger.info({ key: pr._key }, 'PR already exists in job state.');
+            } else {
+              prEntity = (await jobState.addEntity(pr)) as PullRequestEntity;
+            }
```
VDubber, line 104: It seems like we should introduce a `addEntity` that disables dup key check.

VDubber (approved):

src/steps/pullRequests.ts
adam-in-ict: 100% agree.  I've been quietly campaigning for functionality similar to the Cypher MERGE command that does an update/insert based on existence.

I think the trick if we add it will be trying to nudge developers in a good direction so *everything* doesn't become a merge unnecessarily.

VDubber: Maybe if this need functions gets called more than 100 times it blows up or something? :)

PR #210, opened 2022-06-27
Skipping potentially duplicate relationship creation

adam-in-ict:
Skipping both the entity creation and relationship creation if we get a duplicate PR.

### DIFF
--- a/src/steps/pullRequests.ts
+++ b/src/steps/pullRequests.ts
@@ -100,84 +100,85 @@ export async function fetchPrs(
             // cannot have duplicate keys for either.
             if (jobState.hasKey(pr._key)) {
               logger.info({ key: pr._key }, 'PR already exists in job state.');
-            } else {
-              const prEntity = (await jobState.addEntity(
-                pr,
-              )) as PullRequestEntity;
+              return;
+            }
 
+            const prEntity = (await jobState.addEntity(
+              pr,
+            )) as PullRequestEntity;
+
+            await jobState.addRelationship(
+              createDirectRelationship({
+                _class: RelationshipClass.HAS,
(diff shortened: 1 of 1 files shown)


src/steps/pullRequests.ts
```diff
@@ -96,85 +96,88 @@ export async function fetchPrs(
...
-            }
-
-            await jobState.addRelationship(
-              createDirectRelationship({
-                _class: RelationshipClass.HAS,
-                from: repoEntity,
-                to: prEntity,
-              }),
-            );
+              const prEntity = (await jobState.addEntity(
```
VDubber, line 104: nit: Consider adding a return statement in the `if` block to avoid nesting a lot of code in the else block.

src/steps/pullRequests.ts
```diff
@@ -96,85 +96,88 @@ export async function fetchPrs(
             );
             // If we receive a new PR into a repo while paginating, the
             // results will shift and cause us to see a PR twice.
-            let prEntity = await jobState.findEntity(pr._key);
-            if (prEntity) {
+            // We should skip both entity and relationship creation as we
+            // cannot have duplicate keys for either.
+            if (jobState.hasKey(pr._key)) {
               logger.info({ key: pr._key }, 'PR already exists in job state.');
```
VDubber, line 102: Consider improving our confidence that we truly are seeing duplicates by comparing the existing and duplicate entities.

VDubber (approved):

src/steps/pullRequests.ts
adam-in-ict: Thanks!  Don't know why I didn't think of that before. Much more readable.

PR #212, opened 2022-07-01
Update integration-sdk-* to 8.18.0

VDubber:
- Add disabled reasons to getStepStartStates.ts

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -96,7 +96,12 @@ export class APIClient {
 
     const meta = await this.accountClient.fetchMeta();
     this.gheServerVersion = meta.installed_version ?? null;
-    this.logger.info({ meta }, 'API meta');
+    if (this.gheServerVersion) {
+      this.logger.info(
+        { gheServerVersion: this.gheServerVersion },
+        'GitHub Enterprise Server version',
+      );
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -96,7 +96,12 @@ export class APIClient {
 
     const meta = await this.accountClient.fetchMeta();
     this.gheServerVersion = meta.installed_version ?? null;
-    this.logger.info({ meta }, 'API meta');
```
VDubber, line 99: This was logging a large amount of information that wasn't important, especially when the instance was cloud.

src/getStepStartStates.ts
```diff
@@ -19,18 +20,37 @@ export default async function getStepStartStates(
...
+      disabledReason: DisabledStepReason.PERMISSION,
+    },
+    ['fetch-org-secrets']: {
+      disabled: !scopes.orgSecrets,
+      disabledReason: DisabledStepReason.PERMISSION,
+    },
+    ['fetch-repo-secrets']: {
+      disabled: !scopes.repoSecrets,
+      disabledReason: DisabledStepReason.PERMISSION,
+    },
```
adam-in-ict, line 42: So on these if the scope exists and `disabled` is set to false, it'll just ignore the disabledReason, right?

adam-in-ict (approved):

src/getStepStartStates.ts
VDubber: Correct.

PR #214, opened 2022-07-13
Int 4497 Configurable Ingestion start datetime for Pull Requests

VDubber:


.env.example
```diff
@@ -2,4 +2,7 @@ GITHUB_APP_ID=1234
 GITHUB_APP_LOCAL_PRIVATE_KEY_PATH={YOURPATH}/{YOURFILENAME}.private-key.pem
 INSTALLATION_ID=56789
 GITHUB_API_BASE_URL=https://api.github.com
-ENABLE_DEPENDABOT_ALERTS=true
\ No newline at end of file
+
+# Options below are not directly available to the user
```
adam-in-ict, line 6: Good call out, but do you want to specify that you mean through the J1 frontend?  These options are directly available to users running manually or via Starbase still, right?

adam-in-ict (approved):

PR #219, opened 2022-08-19
Int 4497 configurable max pr ingestion

VDubber:
This should fix the issue that carta was experiencing.


adam-in-ict (dismissed):
nit:  we could update the .env.example with this new value, but no need to hold up merging.

adam-in-ict (approved):

PR #224, opened 2022-09-14
Int 415 repoAdmin permission fix

VDubber:


zemberdotnet (dismissed):

src/client.ts
```diff
@@ -688,6 +688,16 @@ export class APIClient {
       this.scopes.dependabotAlerts = false;
     }
 
+    //ingesting branch protection rules requires scope repo administration:read
+    if (['read', 'write'].includes(perms.administration!)) {
```
zemberdotnet, line 692: Why include `write`?

adam-in-ict (approved):

src/client.ts
VDubber: In case a customer sets up their own app and gives write instead of just read. GH Enterprise Serve use case

zemberdotnet (dismissed):

adam-in-ict (approved):

zemberdotnet (approved):

PR #226, opened 2022-09-16
Add Associated Merge Pull Request relationship

VDubber:
Give branches and PRs that build on each other: 

 PR1 - Branch A -> main with commit {A}
 PR2 - Branch B -> main with commits {A, B}
 PR3 - Branch C -> main with commits {A, B, C}

If PR3 is merged first, PR1 and PR2 will be marked as MERGED and potentially without approval. This has caused #security alerts. 

This new CONTAINS relationship will indicate that PR3 ->CONTAINS-> PR1 & PR2

--- 

Open to a different approach to solving the issue.
I went with a relationship instead of marking the merge PR with an indirect approval because once marked, more information would be wanted which resides in the associated PR.
The relationship will only be present after merge.

---

Here's the query that Kenneth was using:
```
FIND PR WITH approved!=true AND state="MERGED"
THAT HAS CodeRepo WITH displayName="login"
```

It'll need to change to something like:
```
FIND PR WITH approved!=true AND state="MERGED"
THAT CONTAINS PR WITH approved!=true and state="MERGED" 
```


src/client/GraphQLClient/fragments.ts
```diff
@@ -114,6 +114,16 @@ export default {
       }
     }
   }`,
+  associatedPullRequest: `
+    associatedPullRequests(first: 1) {
```
adam-in-ict, line 118: Are we deliberately setting this to only pull in a single associated pull request?

src/client/GraphQLClient/pullRequestQueries/__snapshots__/PullRequestsQuery.test.ts.snap
```diff
@@ -254,7 +254,17 @@ Object {
       }
     }
   }
+
```
adam-in-ict, line 257: Would we want to use the defined fragment for associated pull requests here instead? (also in a few additional spots, but I haven't flagged them yet in case I'm wrong).

src/client/GraphQLClient/fragments.ts
VDubber: Yes, in context, the parent of the associatedPullRequest is the merge commit of the PR. 
From what I have tested, the commit and PR are tied together. 
I could totally be wrong here.

adam-in-ict: AH.  So this query is finding from the child PR to the parent PR, not the other way around.  That's what was tripping me up.  I was thinking this was parent to (many) child PRs and we'd miss some.  Thanks!

adam-in-ict (approved):

PR #232, opened 2022-10-31
Include `Discussion` permission to fix BRP Ingestion

VDubber:


electricgull (approved):

src/getStepStartStates.ts
```diff
@@ -53,7 +53,7 @@ export default async function getStepStartStates(
         : DisabledStepReason.PERMISSION,
     },
     ['fetch-branch-protection-rules']: {
-      disabled: !scopes.repoAdmin,
+      disabled: !scopes.repoAdmin && !scopes.repoDiscussions,
```
adam-in-ict, line 56: Is there any concern that this will disable branch protection rule ingestion for users who only care about ingesting rules for public repos and therefore opted not to provide repoDiscussions permissions?

VDubber: Double checking my logic...
This was my goal:
If both scopes are not provided, disable the step. If either scope is provided, do not disabled the step.

adam-in-ict: Oops.  Yeah, you're correct.  Should've double checked my understanding of the logic (and the unit tests).  Thanks!

adam-in-ict (approved):

adam-in-ict (approved):

PR #241, opened 2023-02-14
fix: remove query fragments from versions that do not support them.

RonaldEAM:
From Github GraphQL API changelog:
In version 3.6.0
- Member `App` was added to Union type BranchActorAllowanceActor (used in `bypassForcePushAllowances` and `bypassPullRequestAllowances`)
- Member `App` was added to Union type ReviewDismissalAllowanceActor(used in `reviewDismissalAllowances`)

In version 3.5.0
- Field `blocksCreations` was added to object type `BranchProtectionRule`

This PR removes those fields from the query for versions below these.

Note: `pushAllowances` has members `App`, `Team` and `User` in all versions

### DIFF
--- a/.github/workflows/peril.yml
+++ b/.github/workflows/peril.yml
@@ -1,9 +1,9 @@
-name: "Peril"
+name: 'Peril'
 
 on:
   pull_request:
 
-env: 
+env:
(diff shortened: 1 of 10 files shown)


VDubber (approved):
Awesome!

PR #252, opened 2023-05-01
IDEA-950 - Add repo visibility to `github_repo`

austinkelleher:
| Property     | Type     |
| ------------ | -------- |
| `visibility` | `string` |

### DIFF
--- a/jupiterone/questions/questions.yaml
+++ b/jupiterone/questions/questions.yaml
@@ -1,162 +1,175 @@
 ---
 sourceId: managed:github
-integrationDefinitionId: "${integration_definition_id}"
+integrationDefinitionId: '${integration_definition_id}'
 questions:
-- id: integration-question-github-access-repo-outside-collaborator
-  title: Which GitHub repos are accessible to outside collaborators?
-  description:
-    Finds all GitHub repos accessible to outside collaborators
-  queries:
-  - query: |
-      FIND github_repo AS repo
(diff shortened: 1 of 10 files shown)


src/client/GraphQLClient/fragments.ts
```diff
@@ -83,6 +83,7 @@ export default {
     pushedAt
     rebaseMergeAllowed
     url
+    visibility
```
austinkelleher, line 86: Here was where the new property was added in the GraphQL request

src/sync/converters.ts
```diff
@@ -403,6 +403,7 @@ export function toRepositoryEntity(data: OrgRepoQueryResponse): RepoEntity {
     lockReason: data.lockReason || '',
     mergeCommitAllowed: data.mergeCommitAllowed,
     rebaseMergeAllowed: data.rebaseMergeAllowed,
+    visibility: data.visibility,
```
austinkelleher, line 406: Converter update

ndowmon (dismissed):
❤️

jupiterone/questions/questions.yaml
```diff
@@ -1,162 +1,175 @@
 ---
 sourceId: managed:github
-integrationDefinitionId: "${integration_definition_id}"
+integrationDefinitionId: '${integration_definition_id}'
```
austinkelleher, line 3: All of this was prettier formatting...

PR #256, opened 2023-06-05
Create Packages from Integration

jmountifield:
# Description

This introduces a GitHub workflow to create a signed GitHub package from the integration upon tag push. This is achieved though a Dockerfile and a package-publish action. The signing is keyless signing using cosign.

The integration code itself is unchanged, and has not been tested when running in the produced container.


Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:14-alpine
```
adam-in-ict, line 1: I think this integration has been upgraded to node version 18 already.

jmountifield: Should be resolved now

adam-in-ict (approved):

PR #260, opened 2023-07-03
Collector package deployment step update for E2E tests

maxheckel:


docs/jupiterone.md
```diff
@@ -163,7 +163,6 @@ The following entities are created:
 | GitHub Issue                   | `github_issue`                  | `Issue`         |
 | GitHub Org Secret              | `github_org_secret`             | `Secret`        |
 | GitHub Pull Request            | `github_pullrequest`            | `PR`            |
-| GitHub Outside Collaborator    | `github_user`                   | `User`          |
```
adam-in-ict, line 166: Should this be removed?

maxheckel: It was an auto-generated change, so I guess so?

zemberdotnet (approved):

docs/jupiterone.md
VDubber: I believe this is a bug in the docs generator. This line should be reverted

PR #267, opened 2023-07-28
[INT-8773] - Refactor validate invocation

dobregons:
## Summary
According to the `integration-sdk-core` package, the `validateInvocation` should not return any type, instead, it should return void. https://github.com/JupiterOne/sdk/blob/main/packages/integration-sdk-core/src/types/validation.ts#L4-L6

The previous version of `validateInvocation` didn't follow this interface which prevented returning this function as part of the `invocationConfig`. This PR fixes that problem and create another function to be used in `getStepStartStates` that validates the invocation and return the authentication data needed there. That's the only place where we need the validateInvocation with the return authentication data.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,6 +4,7 @@ import {
   IntegrationInstanceConfigFieldMap,
   IntegrationInstanceConfig,
   IntegrationIngestionConfigFieldMap,
+  IntegrationLogger,
 } from '@jupiterone/integration-sdk-core';
 import { getOrCreateApiClient } from './client';
 const fs = require('fs');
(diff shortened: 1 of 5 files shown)


src/config.ts
```diff
@@ -139,6 +141,22 @@ export async function validateInvocation(
   };
 }
 
+export async function validateInvocation(
+  context: IntegrationExecutionContext<IntegrationConfig>,
+): Promise<void> {
+  const { instance, logger } = context;
```
VDubber, line 147: consider calling `validateAndReturnAuthenticationData` without a return statement.

VDubber (approved):

PR #268, opened 2023-07-28
INT-8436 - Use external identifiers for better email resolution

adam-in-ict:
Adding an enrichment step to the `Fetch Users` step to optionally query for external identifiers for organizations that have enabled such.  Based on this information, we can resolve additional organizational email addresses for use in mapped relationships later.


src/sync/converters.ts
```diff
@@ -429,7 +430,7 @@ export function toOrganizationMemberEntity(
     login: data.login,
     username: data.login,
     displayName: data.name || data.login,
-    name: data.name,
+    name: data.name || data.login,
```
adam-in-ict, line 433: Bonus change.  We're not guaranteed to have a value for data.name, so falling back to the same value as we're using in displayName.

gastonyelmini (dismissed):
Nice!

gastonyelmini (dismissed):

gastonyelmini (approved):

PR #294, opened 2024-02-23
Add authentication type for enterprise PATs

RonaldEAM:
- refactored clients into only two separate ones, RestClient and GraphqlClient
- added new authentication type for enterprise owner's personal access token
- ingest external identities from enterprise instead of organization when using the new authentication


Gonzalo-Avalos-Ribas (approved):
LGTM!

PR #299, opened 2024-03-20
skip empty pull requests response object

RonaldEAM:


src/client/GraphQLClient/pullRequestQueries/PullRequestsQuery.ts
```diff
@@ -120,7 +120,7 @@ export const processResponseData = async (
 
   return {
     rateLimit,
-    pullRequests: responseData.repository.pullRequests.pageInfo,
+    pullRequests: responseData.repository?.pullRequests?.pageInfo,
```
VDubber, line 123: nit: I would expect a var named `pullRequests` to have more than just pageInfo.

VDubber (approved):

src/client/GraphQLClient/pullRequestQueries/PullRequestsQuery.ts
RonaldEAM: yes I was just thinking that... but it's a pattern used in all other queries, I'll fix it in another PR for all for them

PR #301, opened 2024-05-14
Fix tests dates

Nick-NCSU:


tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "extends": "./node_modules/@jupiterone/integration-sdk-dev-tools/config/typescript",
+  "allowSyntheticDefaultImports": true,
```
Nick-NCSU, line 3: Allows import of the utc plugin

src/steps/pullRequests.ts
```diff
@@ -327,7 +328,8 @@ export const determineIngestStartDatetime = (
     return dayjs(config.pullRequestIngestStartDatetime).toISOString();
   }
 
-  const nowDate = dayjs();
+  dayjs.extend(utc);
+  const nowDate = dayjs().utc();
```
Nick-NCSU, line 332: We now always use UTC when getting the date. It is no longer dependent on the system timezone.

src/steps/pullRequests.test.ts
```diff
@@ -17,8 +17,13 @@ import MockDate from 'mockdate';
 const filterOutPrContainsPrRelationships = (r: any) =>
   r._type !== Relationships.PULLREQUEST_CONTAINS_PULLREQUEST._type;
 
+beforeEach(() => {
+  MockDate.set('2023-12-31T12:00:00.000Z');
+});
```
Nick-NCSU, line 22: The `fetchPrs` was created a little over 365 days ago so the recording expired. Moved the mock date to run before all tests so that it cannot expire.

i5o (approved):

